Issue #36, opened 2015-08-19
don't use ssh to clone repos in scripts (use https)


elf-pavlik:
* https://github.com/inventaire/inventaire/blob/master/scripts/install-client#L1
* https://github.com/inventaire/inventaire/blob/master/scripts/postinstall#L6

maxlath:
isn't it a question of preferences? I do prefer to use ssh as then I don't have to enter my password again while https would ask for it everytimes I try to push something

elf-pavlik:
I don't think anyone else than you will push to this branch, and you can easily change git remote from http to ssh in your local clone... Do you use ssh key not protected with a pass password?

`git remote set-url origin git@github.com:maxlath/inventaire-client.git`
http://git-scm.com/docs/git-remote.html

frimelle:
I agree.
There shouldn't be something that hinders any one on installing it like a personalization in the ssh.

PR #50, opened 2015-10-17
Use https in scripts

frimelle:
Resolve inventaire/inventaire#36

### DIFF
--- a/scripts/install-client
+++ b/scripts/install-client
@@ -1,3 +1,3 @@
-git clone git@github.com:maxlath/inventaire-client.git ./client
+git clone https://github.com/inventaire/inventaire-client.git ./client
 cd client
 npm install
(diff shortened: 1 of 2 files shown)


maxlath:
my fault! I should indeed keep the monopoly on pushing to master for now, thank you for creating this new situation and helping me to learn :D 
I rolled back to the previous state using a [forced push](http://stackoverflow.com/a/6872462/3324977), you might need to delete your local master branch and re-pull it to be back in sync.
could you make a pull request with only 68fc461? it would make the history cleaner :)

maxlath:
so in the end, to keep it clean, I cherry-picked your 68fc461 commit, which thus became 0c7d575
Thanks for your PR: you've earned the title of first non-me human contributing to this repo!! \o/

frimelle:
Yeah! I'll take that honour with proud ;)

PR #95, opened 2017-06-06
Rename rank to ordinal

thewholelifetolearn:
Rename 'rank' to 'ordinal' to be consistent with Wikidata


maxlath:
thanks, and congrats for your first PRs here \o/

thewholelifetolearn:
I hadn't thought about the queries. But I didn't see those files modified lately. I checked your recent commits but I didn't see anything about queries.

maxlath:
that's normal, as those queries where simply relying on `wdt:P1545` to order series parts, but without anything else relying on it or displaying the corresponding ordinal:  it wouldn't have break anything to keep those as they where before 537acbf, but if we can be consistent in the naming, that's all better

PR #109, opened 2017-10-29
102 recent public items

jum-s:
Todo:

- on the client : call `recent-public` instead of  `last-public`
- (maybe) extract [those](https://github.com/inventaire/inventaire/compare/master...102-recent-public-items#diff-c0ac6bb71687269dff5cd3eaa9f21cc4R18-19) variables outside function


api_tests/items/recent_public_items.coffee
```diff
@@ -0,0 +1,35 @@
...
+
+describe 'items:recent-public', ->
+  it 'should fetch 15 recent-public items', (done)->
+    populate
+      usersCount: 1
+      publicItemsPerUser: 16
+    .then -> nonAuthReq 'get', recentPublicUrl
+    .then (res)->
+      res.items.length.should.equal 15
+
```
maxlath: style note: the lines 15-17 could be turned into a one-liner like the line 14

server/controllers/items/recent_public.coffee
```diff
@@ -0,0 +1,40 @@
...
+  maxItemsPerOwner = 3
+
+  Promise.resolve _.groupBy items, (item)-> item.owner
+  .then (itemsByOwner)->
+    _.map itemsByOwner, (items)->
+      items.slice 0, maxItemsPerOwner
+  .then (limitedItemsByOwner)->
+    _.flatten _.values limitedItemsByOwner
+  .then (allRecentItems)->
+    allRecentItems.slice 0, itemsLimit
```
maxlath: lines 21-28 could be refactored with [Lodash sequence methods](https://lodash.com/docs/#lodash) as
```js
  _(items)
  .groupBy _.property('owner')
  .map (items)-> items.slice(0, maxItemsPerOwner)
  .values()
  .flatten()
  .slice 0, itemsLimit
  .value()
```

server/controllers/items/recent_public.coffee
```diff
@@ -0,0 +1,40 @@
...
+
+bundleOwnersData = (res, reqUserId, items)->
```
maxlath: for functions that are duplicated with [./last_public](https://github.com/inventaire/inventaire/blob/0cf55dc/server/controllers/items/last_public.coffee), you can either extract the concerned functions to a third module both controllers would require, or we could consider that `recent-public` makes `last-public` obsolete and `git rm` it

server/controllers/items/recent_public.coffee
```diff
@@ -0,0 +1,40 @@
...
+  reqUserId = req.user?._id
+
+  items_.publicByDate 100, 0, assertImage, reqUserId
+  .then selectRecentItems
+  .then bundleOwnersData.bind(null, res, reqUserId)
+  .catch error_.Handler(req, res)
+
+selectRecentItems = (items)->
+  itemsLimit = 15
+  maxItemsPerOwner = 3
```
maxlath: this kind of hardcoded parameters could be defined just below the `require`s

api_tests/items/recent_public_items.coffee
jum-s: readability concern, as `.should` should is the first things to read in this file

server/controllers/items/recent_public.coffee
jum-s: as `last_public` will not be used, dead code can be removed
git archives it anyway

server/controllers/items/recent_public.coffee
```diff
@@ -15,17 +15,23 @@ module.exports = (req, res)->
...
-    allRecentItems.slice 0, itemsLimit
+  _(items)
+  .groupBy itemOwner
+  .map (items, _)-> items.slice(0, maxItemsPerOwner)
+  .flatten()
+  .values()
+  .slice 0, itemsLimit
+  .shuffle()
+  .value() # wrapping lodash .chain() function
+
```
maxlath: In this kind of chain, I generally prefer to have the comments directly on top of what then comment:
```coffee
  _(items)
  # Group items in an object of owners ID,
  .groupBy itemOwner
  # then select only x firsts items from each owner
  .map (items, _)-> items.slice(0, maxItemsPerOwner)
  # turn object into flattened array
  .flatten()
  .values()
  .slice 0, itemsLimit
  .shuffle()
  # wrapping lodash .chain() function
  .value()
```
examples: [1](https://github.com/inventaire/inventaire/blob/5b9bc9d/server/controllers/invitations/lib/parse_emails.coffee#L22-L28), [2](https://github.com/inventaire/inventaire/blob/5b9bc9d/server/controllers/groups/lib/slugify.coffee), [3](https://github.com/inventaire/inventaire/blob/5b9bc9d/server/controllers/entities/lib/scaffold_entity_from_seed/search_work_entity_by_title_and_authors.coffee#L32-L37)

server/controllers/items/items.coffee
```diff
@@ -11,7 +11,6 @@ module.exports =
       'by-ids': require './by_ids'
       'by-users': require './by_users'
       'by-entities': require './by_entities'
-      'last-public': require './last_public'
```
maxlath: we actually still need this endpoint for the `last public` inventory layout
last_public
It might be made obsolete by the incoming [inventory browser](https://github.com/inventaire/inventaire-client/tree/inventoryBrowser) view, but it's too early to tell, so better let it around for now

server/controllers/items/recent_public.coffee
maxlath: or actually, they would probably be better defined dynamically from the query string (like how the `limit` is defined for the `last-public` action), so that the client can tell how many items it want, depending on the screen width

maxlath:
once [this PR](https://github.com/inventaire/inventaire/pull/115) is merge, you could rebase on top of it to take advantage of the `populate` speed improvements

server/controllers/items/last_public.coffee
```diff
@@ -0,0 +1,43 @@
...
+    return error_.bundleInvalid req, res, 'offset', offset
+
+  if limit > 100
+    return error_.bundle req, res, "limit can't be over 100", 400, limit
+
+  items_.publicByDate limit, offset, assertImage, reqUserId
+  .then bundleOwnersData.bind(null, res, reqUserId)
+  .catch error_.Handler(req, res)
+
+bundleOwnersData = (res, reqUserId, items)->
```
maxlath: This should be DRYed with `recent-public` module as it's exactly the same code. It can be done by creating a third module in `server/controllers/items/lib`

jum-s:
rebased, ready to merge

server/controllers/items/lib/items.coffee
```diff
@@ -146,6 +146,17 @@ module.exports = items_ =
     if image? and item.pictures.length is 0 then item.pictures = [ image ]
     return item
 
+  bundleOwnersData: (res, reqUserId, items)->
+    unless items?.length > 0
```
maxlath: L149-159: this can be removed now that both server/controllers/items/recent_public.coffee and server/controllers/items/last_public.coffee use `bundleOwnersToItems = require './lib/bundle_owners_to_items'`

maxlath:
almost :D

PR #111, opened 2017-10-29
crypto: allowing to disable password hashing

maxlath:
to make signup run faster in test environment


jum-s:
belle! thx for the detailed commenting part, helps a lot.
for the record, this was implemented after the [testFixtures](https://github.com/inventaire/inventaire/pull/108) populate in order to create user faster in a test env

PR #113, opened 2017-10-29
Adding scripts to generate NDJSON and TTL dumps of the entity database

maxlath:
to be published on dumps.inventaire.io. Also unlocking the possibility to load a dump in a BlazeGraph instance


maxlath:
@jum-s review corrections done in 199d6a4

PR #115, opened 2017-11-05
added bulk item creation

maxlath:
and make use of it to speed-up fixture `populate` function

### DIFF
--- a/api_tests/fixtures/populate.coffee
+++ b/api_tests/fixtures/populate.coffee
@@ -3,21 +3,21 @@ __ = CONFIG.universalPath
 _ = __.require 'builders', 'utils'
 { Promise } = __.require 'lib', 'promises'
 { createUser } = require './users'
-{ createRandomizedItem } = require './items'
+{ createRandomizedItems } = require './items'
+randomString = __.require 'lib', './utils/random_string'
 
-module.exports = (params) ->
-  { usersCount } = params
-  promiseArray = [1..usersCount].map ->
-    createUserWithItems(params)
-  Promise.all promiseArray
+populatePromise = null
+usersCount = 2
+publicItemsPerUser = 16
 
-createUserWithItems = (params) ->
-  { itemsPerUser, publicItemsPerUser } = params
+module.exports = ->
+  if populatePromise? then return populatePromise
+  populatePromise = Promise.all _.times(usersCount, createUserWithItems)
+  return populatePromise
+
+createUserWithItems = ->
   userPromise = createUser()
-  userPromise.then ->
-    if itemsPerUser
-      [1..itemsPerUser].forEach ->
-        createRandomizedItem userPromise
-    if publicItemsPerUser
-      [1..publicItemsPerUser].forEach ->
-        createRandomizedItem userPromise, { listing: 'public'}
+  userPromise
+  .then ->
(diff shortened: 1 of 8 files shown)


api_tests/fixtures/populate.coffee
```diff
@@ -3,21 +3,21 @@ __ = CONFIG.universalPath
...
+randomString = __.require 'lib', './utils/random_string'
 
-module.exports = (params) ->
-  { usersCount } = params
-  promiseArray = [1..usersCount].map ->
-    createUserWithItems(params)
-  Promise.all promiseArray
+populatePromise = null
+usersCount = 2
+publicItemsPerUser = 16
```
jum-s: maybe 3 items per user by default is enough, less than 16 would be better

server/controllers/items/lib/items.coffee
```diff
@@ -66,11 +66,15 @@ module.exports = items_ =
     db.viewByKeys 'byOwnerAndEntityAndListing', keys
     .map filterPrivateAttributes(reqUserId)
 
-  create: (userId, item)->
-    db.post Item.create(userId, item)
+  create: (userId, items)->
+    _.type items, 'array'
+    Promise.all items.map(snapshotEntityData)
```
jum-s: it would be a great opportunity to describe/document what `snapshotEntityData` is doing, cuz its a bit cryptic and its fundamental to understand the review  :)
maybe at the beginning of the file  `items/lib/snapshot/snapshot_entity_data.coffee` directly ?

server/controllers/items/create.coffee
```diff
@@ -3,24 +3,32 @@ _ = __.require 'builders', 'utils'
...
-  { entity:entityUri } = item
-  unless entityUri? then return error_.bundleMissingBody req, res, 'entity'
+  items = _.forceArray items
 
-  unless _.isEntityUri entityUri
-    return error_.bundleInvalid req, res, 'entity', entityUri
+  _.log items, 'create items'
+
+  for item in items
+    { entity:entityUri } = item
```
jum-s: new syntax for me.
meaning `entityUri = item[entity]` ?

api_tests/fixtures/populate.coffee
maxlath: I choose `2` and `16` to match the parameters from [`recent_public_items`](https://github.com/inventaire/inventaire/pull/109/files#diff-2882db99c0bf7b4963427781e781053b) tests, as the populate function doesn't take parameters anymore, but once this branch is merge, you are welcome to rebase your branch on top of it and customize it to fit your tests

server/controllers/items/create.coffee
maxlath: pretty much (`entityUri = item.entity` or `entityUri = item['entity']` to be exact)
It's part of the new ES6 [destructuring assignment](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#Assigning_to_new_variable_names) capabilities that were ported back in CoffeeScript

server/controllers/items/lib/items.coffee
maxlath: yes! done in 55d6127

Issue #127, opened 2017-11-19
allow to request recent public items filtered by language


maxlath:
the current welcome page displays a list of books taken from `/api/items?action=recent-public`, which might give the impression that there are only books in other language. It could be more inviting to be displayed books in the language matching the user language.

Issue #120, opened 2017-11-06
should we merge the server and client repositories?


maxlath:
**Pros**:
- shared libs edits are clearer: currently, a piece of code moved from the server to the shared libs just disappears from the server, in the best cases with a link to the corresponding commit in the client repo
- features with strong coupling of server and client code should make more sense

**Cons**:
- having everything in one repo might make the commit history harder to browse

In the case we choose to merge:
- [How do you merge two Git repositories?](https://stackoverflow.com/a/10548919/3324977)

jum-s:
caution context: only talking of git here, decoupling server and client code is another topic

Today, client and server live on the same production server, pushed from the same repo.
when a new server branch that creates a new feature is merged, a client commit usually follows on a client repo, that ends up being linked through comments like :

![selection_009](https://user-images.githubusercontent.com/5363918/33240335-d44ddd80-d2b4-11e7-8f5b-464fa021b35a.png)

Having a separated git history can benefit bisecting, and forking (harder to browse can be fixed by branching and stacking). But a client is not built overnight anyway, when a new client emerge we will ask each other question, but so far that doesnt ease the documentation of the project.

maxlath:
a tool that could help: [git-merge-repos](https://github.com/robinst/git-merge-repos)

Issue #72, opened 2016-10-13
from coffeescript to ES6+


maxlath:
Inventaire started as [a love affair with coffeescript](https://github.com/inventaire/inventaire/wiki/CoffeeScript) but it's not a definitive thing. While I don't mind using technologies that were trendy more than 6 months ago, there are real incentives to move the whole project to standard JS, ES6 and above.
## Reasons to stay with CS
- on the client side, **you need a transpiler anyway**. Browsers are not ready yet for pure ES6, and when they will be, they won't be ready for ES7, etc.
- **cleaner code** thank to meaningful indentations, beautiful [switch statements](https://github.com/maxlath/wikidata-sdk/blob/master/src/helpers/simplify_claims.coffee#L31-L38), comma-less/bracket-less objects (look at those [factory functions](https://github.com/inventaire/inventaire/blob/dev/server/controllers/relations/lib/parse_relations.coffee#L7-L11) <3), clean loops
- the existence operator `?`
- ~~CS is currently the fastest way for this project main contributor (hello :)) to get his ideas into code. But that's a vicious circle isn't it?~~
- ~~moving to ES6 modules would make [`universalPath`](https://github.com/inventaire/inventaire/blob/master/config/universal_path.coffee) impossible (see [ES6 variable import name in node.js](http://stackoverflow.com/questions/29168433/es6-variable-import-name-in-node-js))~~ *actually, we wouldn't be forced to use the ES6 module syntax, the CommonJS flavor would still be an option*
- arrow functions that aren't binding the context and giving the `arguments` object that [horrible/inconsistent behavior ES6 arrow functions](http://stackoverflow.com/questions/33288998/do-es6-arrow-functions-have-their-own-arguments-or-not) give
## Reasons to move to ES6+
- **Having a language known by way more people**. Despite the fact that _CS is just JS_, it isn't obvious to the new comers, making it less likely for them to jump in the code
- **Not having a language denigrated by the whole community**. Yep, that's sad, but new learners are largely discouraged to learn CS by the community, and while one might think this is totally disproportionate, you got to pick your fights
- With such a bad press, we can expect less and less support for CS, in JS tools in general (that `--coffee` flag in test tools, etc), but also in CS specific tools (coffeelint)
- leave CS kind of toxic governance for the TC39 bunch of java-lovers... hurraaay... :p
- NodeJs is ready for ES6: yeah destructive assignments \o/
[...]

PR #128, opened 2017-11-23
127 recent public per language

maxlath:
allow to request recent public items filtered by language #127
\+ some cleaning of the `/api/items?action=recent-public` endpoint


server/controllers/items/last_public.coffee
```diff
@@ -7,7 +7,7 @@ bundleOwnersToItems = require './lib/bundle_owners_to_items'
 module.exports = (req, res)->
   { query } = req
   { limit, offset } = query
-  assertImage = query['assert-image'] is 'true'
+  assertImage = _.parseBooleanString query['assert-image']
```
jum-s: why lodash more than vanilla ? to me lodash is for complex manipulation, this shoudl stay as it was

server/controllers/items/lib/bundle_owners_to_items.coffee
```diff
@@ -2,6 +2,7 @@ CONFIG = require 'config'
 __ = CONFIG.universalPath
 _ = __.require('builders', 'utils')
 user_ = __.require 'controllers', 'user/lib/user'
+error_ = __.require 'lib', 'error/error'
```
jum-s: ive seen that elsewhere (dont remember). why is it needed if not called?

server/controllers/items/lib/items.coffee
```diff
@@ -60,6 +60,17 @@ module.exports = items_ =
     .then FilterWithImage(assertImage)
     .map filterPrivateAttributes(reqUserId)
 
+  publicByLangAndDate: (limit, offset, lang, assertImage, reqUserId)->
```
jum-s: duplictation with `publicByDate` which is only used twice in the system, dont you think we can delete it and integrate it within `publicByLangAndDate`

server/controllers/items/recent_public.coffee
```diff
@@ -3,29 +3,42 @@ _ = __.require 'builders', 'utils'
...
+
+  lang or= 'en'
+
+  assertImage = _.parseBooleanString query['assert-image']
+  reqUserId = req.user?._id
 
-  items_.publicByDate itemsQueryLimit, offset, assertImage, reqUserId
-  .then selectRecentItems.bind(null, itemsReturnedLimit, maxItemsPerOwner)
+  items_.publicByLangAndDate itemsQueryLimit, offset, lang, assertImage, reqUserId
+  .then selectRecentItems(limit, maxItemsPerOwner)
```
jum-s: why changing the limit name ? `itemsReturnedLimit` is supposed to make a clear distinction with the other limit of this function `itemsQueryLimit`

server/controllers/items/recent_public.coffee
```diff
@@ -3,29 +3,42 @@ _ = __.require 'builders', 'utils'
...
-  offset = 0
-  itemsReturnedLimit = 15
-  maxItemsPerOwner = 3
+  if lang? and not _.isLang(lang)
+    return error_.bundleInvalid req, res, 'lang', lang
+
+  lang or= 'en'
+
+  assertImage = _.parseBooleanString query['assert-image']
+  reqUserId = req.user?._id
```
jum-s: its painful to see how the typing/validation part is sooooo long (readability is smashed just to check that limit is below 100...) seriously ?! JS isnt capable of doing better that? can we have an easier convention ? i dont know any easy solution ? cant we switch to typescript ? anyway my eyes bleed can we talk before merge ? :)

server/controllers/items/last_public.coffee
maxlath: there's a confusion there: `parseBooleanString` isn't defined by lodash but as a [shared utils](https://github.com/inventaire/inventaire-client/blob/09d553d/app/lib/shared/utils.coffee#L74).
[Many utils functions are added on `_`](https://github.com/inventaire/inventaire/blob/b93c9c4/server/builders/utils.coffee#L16) without implementing Lodash chaining features or anything, just as a way to access those generalist functions easily, but it might be too much and could desired to be split to ease comprehension of where a given function is defined.

But back in the present case, I'm for using  `parseBooleanString` as, while not doing much, it gives the statement a "this is a common pattern" tag, which can then be searched through the code base more easily

server/controllers/items/lib/bundle_owners_to_items.coffee
maxlath: it is called on line 9

server/controllers/items/lib/items.coffee
maxlath: we could, but in a second movement in my opinion: I tried to think of a view that could be used by both but could find any, so to make `last-public` use the `publicByLangAndDate` view, we need to make it filter by lang too. Which in turn would imply updating the client `/inventory/last` layout. But once we are done with our use case of `recent-public` on the `/welcome` layout, we could have a look at how to merge those two endpoints.

In general, I prefer to start by duplicating, then progressively push everything in direction of the new implementation

server/controllers/items/recent_public.coffee
maxlath: there is probably room for improvement: not knowing better, I kind of went for rewriting my own validation framework (while there are existing express solutions ([example](https://github.com/ctavan/express-validator)), but I consider controllers validation being ugly as acceptable, given it plays the role of interface between API requests (which should get nice, explicit error messages) and internal libraries (which should live in a beautiful world of pre-validated inputs, free of all the trauma of having to take input from the external world)

And I'm no expert, but I don't see how typescript would help here. But since you bring it, the first step to get there is to solve #72 ;)

But yes, you can add it to our next call agenda if this comment doesn't fully address your concern :)

server/controllers/items/recent_public.coffee
maxlath: I'm re-using the parameter's name as it is now passed in the request query. Given the limit is defined by the request, it's purpose to be used as the items returned limit seems quite clear. We could have a short line of comment on the line 30 to explicit why we need to query 100 before selecting, to make the distinction between those 2 limits even more clear

server/controllers/items/lib/bundle_owners_to_items.coffee
jum-s: right ! another untested line...

server/controllers/items/recent_public.coffee
jum-s: this [interface](http://www.typescriptlang.org/docs/handbook/typescript-in-5-minutes.html) thing is appealing to me

```
interface Person {
    firstName: string;
    lastName: string;
}
function greeter(person: Person) {
    return "Hello, " + person.firstName + " " + person.lastName;
}
let user = { firstName: "Jane", lastName: "User" };
```

server/controllers/items/last_public.coffee
maxlath: because that's the first time you encounter `parseBooleanString`, but [next time you will see it](https://github.com/inventaire/inventaire/search?utf8=%E2%9C%93&q=parseBooleanString&type=), the intent will be obvious

jum-s: that doesnt answer my remark, the intent is less obvious 
it also creates a different way to assign it, between `recent_public` and `last_public`, at least change it also on the  `recent_public`.
If the goal is to DRY those two in the future, and we write syntax differences just because "its already somewhere else in the code, so its fine", then this goal is getting a step further.

server/controllers/items/lib/items.coffee
jum-s: "Which in turn would imply updating the client " one more pro for the [git repo merge](https://github.com/inventaire/inventaire/issues/120)
indeed its not that easy (yet) lets see later

server/controllers/items/last_public.coffee
maxlath: that's already the case: [`recent_public` uses it too](https://github.com/inventaire/inventaire/blob/0d14635/server/controllers/items/recent_public.coffee#L27)

jum-s: to be honest, following lodash syntax rather than coffee, converges more toward es6 standards

Issue #85, opened 2017-03-09
split databases per document type


maxlath:
some document types are being mixed in common databases. From [server/db/couch/list.coffee](https://github.com/inventaire/inventaire/blob/2c3ddea28236177d5d04e9fb0b06ce125015a762/server/db/couch/list.coffee):
```js
users: ['user', 'relations', 'groups', 'invited']
entities: ['entities', 'patches']
```

This could [degrade performance at scale](http://www.dimagi.com/blog/what-every-developer-should-know-about-couchdb/), without bringing much value: the original idea was to use [linked documents](http://docs.couchdb.org/en/2.0.0/couchapp/views/joins.html) and ease maintenance (less databases to take care of). This last (weak) point on maintenance is already being counter-balanced by the additional care having several types in a single databases requires when performing a [migration](https://github.com/inventaire/inventaire/blob/2c3ddea28236177d5d04e9fb0b06ce125015a762/scripts/migrator).

maxlath:
done for groups: 2cd3130

maxlath:
done for patches: 9f3b786

jum-s:
done according to [the new list](https://github.com/inventaire/inventaire/blob/master/server/db/couch/list.coffee)  
`patches` & ` groups` has its own type

PR #131, opened 2017-11-27
[WIP] 129 cleaning entities

jum-s:
to discuss ! lots of decision already !

- seedcheck : quid of the unhandy search by text, replace it with `SearchWikidataByText` to ease the testing
- task : new task object attributes naming is clear enough? 
- next step : updating if already existing task ?


server/controllers/tasks/lib/check_seed.coffee
```diff
@@ -0,0 +1,32 @@
...
+    disableDataseed: true
+  .then (entitySearchResult)->
+    return {} unless entitySearchResult
+    entitySearchResult
+    .map (dominantEntity)->
+      dominantUri = dominantEntity.uri
+      if isSubsetOf(claims, dominantEntity.claims)
+        type: "task"
+        possibleDuplicates:
+          [ "#{_id}": "#{dominantUri}" ]
```
jum-s: to be replaced by Task.create

server/models/task.coffee
```diff
@@ -0,0 +1,22 @@
...
+
+  tests.pass 'suspect', suspect
+
+  now = Date.now()
+
+  return task =
+    type: "task"
+    suspectUri: suspect
+    suggestionUri: suggestion
+    state: 'requested'
```
jum-s: as a suggestion for next steps

test/libs/047-entity-collector.coffee
```diff
@@ -0,0 +1,37 @@
...
+__ = CONFIG.universalPath
+_ = __.require 'builders', 'utils'
+
+createHuman = (label)->
+  _id: '91760e03c9c367c3fd75655532f4b2f6'
+  type: "human"
+  labels: { en: label },
+  claims: { 'wdt:P31': [ 'wd:Q5' ] }
+
+checkEntity = __.require 'controllers', 'tasks/lib/check_entity'
```
maxlath: as `checkEntity` is async and non-deterministic (relies on the Wikidata API), those tests should probably be moved to `api_tests`: tests in the `test` folder are run before every commit (see [pre-commit githooks](https://github.com/inventaire/inventaire/blob/master/.githooks/pre-commit/pre-commit#L10)), they should thus run really fast and should fail only in the case something in the code broke the code (not in case of network error or anything of this kind)

server/controllers/tasks/lib/check_entity.coffee
```diff
@@ -0,0 +1,48 @@
...
+  searchWikidataByText
+    search: title
+  .then (searchResult)->
+    entities = _.values(searchResult.entities)
+    filterSuggestions entity, entities
+  .then (entities)->
+    tooManyHomonyms(entities)
+  .then (suggestionEntities)->
+    newTasks = createTasks entity, suggestionEntities
+    return { "tasks": newTasks }
```
maxlath: so now that we have a lib to generate tasks from an entity, what we need is a script that would:
- get entities that it would need to be checked: this could rely on another lib that would take of fetching entities that aren't yet targeted by a task
- pass those entities to `checkEntity` and pass the generated tasks docs to another lib function (probably in `controllers/tasks/lib/tasks.coffee`) to post it in `tasks` database (which btw needs to be added to [database list](https://github.com/inventaire/inventaire/blob/3258a58/server/db/couch/list.coffee))

This script could then be run periodically (as a system cron task maybe?) to feed the tasks database. We would need then to make those tasks accessible (`GET`) and updatable (`PUT`) from the client, by creating an `/api/tasks` endpoint, but we will get to that later.

server/models/task.coffee
```diff
@@ -0,0 +1,22 @@
...
+
+Task.create = (suspect, suggestion)->
+  _.types arguments, [ 'string', 'string' ]
+
+  tests.pass 'suspect', suspect
+
+  now = Date.now()
+
+  return task =
+    type: "task"
```
maxlath: it might be a bit anticipating, but, additionally to this type 'task' (which might actually not be needed if we agree to put only documents of type tasks in the tasks database, which is now the [recommended way to do things](https://github.com/inventaire/inventaire/issues/85)), we could have a subtype to explicitly describe the kind of task we are dealing with - here something with "deduplication". The alternative would be to duck-type the kind of task from the task-specific attributes (here, possilbly `suspectUri` and `suggestionUri`), but expliciting it feels cleaner right?

maxlath:
mistake delete, sorry

maxlath:
@jum-s could you review those last commits to validate a rebase and merge on master? (getting there \o/)

server/lib/sanitize/sanitize.coffee
```diff
@@ -36,6 +36,8 @@ sanitizeParameter = (input, name, config, place, res)->
     if config.optional then return
     else throw error_.newMissing place, name
 
+  if parameter.validatedByModel then return
```
jum-s: if i understand well all this, is it the responsability of sanitization lib to check models validation?

maxlath: the opposite: this is a flag to not try to validate in the sanitization lib. The name could be changed to something more explicit maybe, like `noValidation`

PR #134, opened 2017-12-15
make `GET /api/entities?action=by-uris` accept a `relatives` parameter

maxlath:
make `GET /api/entities?action=by-uris` accept a `relatives` parameter to include entities related to the requested entities by following those entities claims
ex:
- fetch a work by URI, and set `relatives=wdt:P50` to also get its authors entities in the response
- fetch an edition by URI, and set `relatives=wdt:P629|wdt:P50|wdt:P179` to also get the associated works, authors, and series

This feature is currently only needed by in [inventaire/inventaire-client#isbns-importer](https://github.com/inventaire/inventaire-client/tree/isbns-importer), so we should wait for the isbns importer feature to be finished to confirm the need for this `relatives` feature. Unless this feature gets consensually accepted as a nice to have for the API, in which we could keep it despite not having any client code relying on it for now

### DIFF
--- a/api_tests/entities/get.test.coffee
+++ b/api_tests/entities/get.test.coffee
@@ -2,9 +2,10 @@ CONFIG = require 'config'
 __ = CONFIG.universalPath
 _ = __.require 'builders', 'utils'
 should = require 'should'
+{ Promise } = __.require 'lib', 'promises'
 { authReq, nonAuthReq, undesiredErr, undesiredRes } = require '../utils/utils'
 { getByUris } = require '../utils/entities'
-{ ensureEditionExists, createWorkWithAuthor } = require '../fixtures/entities'
+{ ensureEditionExists, createWorkWithAuthor, createEditionWithWorkAuthorAndSerie } = require '../fixtures/entities'
 endpointBase = '/api/entities?action=by-uris&uris='
 workWithAuthorPromise = createWorkWithAuthor()
 
@@ -133,6 +134,34 @@ describe 'entities:get:by-uris', ->
           err.statusCode.should.equal 400
           err.body.status_verbose.should.equal 'invalid relative: wdt:P31'
           done()
+      .catch undesiredErr(done)
+
+      return
+
+    it "should be able to include the works, authors, and series of an edition", (done)->
+      createEditionWithWorkAuthorAndSerie()
+      .get 'uri'
+      .then (editionUri)->
+        relatives = 'relatives=wdt:P50|wdt:P179|wdt:P629'
+        nonAuthReq 'get', "#{endpointBase}#{editionUri}&#{relatives}"
+        .then (res)->
+          edition = res.entities[editionUri]
+          edition.should.be.an.Object()
+
+          workUri = edition.claims['wdt:P629'][0]
+          work = res.entities[workUri]
+          work.should.be.an.Object()
+
+          authorUri = work.claims['wdt:P50'][0]
+          author = res.entities[authorUri]
+          author.should.be.an.Object()
+
+          serieUri = work.claims['wdt:P179'][0]
+          serie = res.entities[serieUri]
+          serie.should.be.an.Object()
+
+          done()
+
       .catch undesiredErr(done)
 
       return
--- a/api_tests/entities/get.test.coffee
+++ b/api_tests/entities/get.test.coffee
@@ -46,62 +46,62 @@ describe 'entities:get:by-uris', ->
     return
 
   describe 'alias URIs', ->
-  it 'should accept twitter URIs', (done)->
-    aliasUri = 'twitter:bouletcorp'
-    getByUris aliasUri
-    .then (res)->
-      { entities, redirects } = res
-      canonicalUri = redirects[aliasUri]
-      canonicalUri.should.equal 'wd:Q1524522'
-      entity = entities[canonicalUri]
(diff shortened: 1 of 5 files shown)


api_tests/entities/get.test.coffee
```diff
@@ -45,62 +48,119 @@ describe 'entities:get:by-uris', ->
...
+  describe 'relatives', ->
+    it "should accept a 'relatives' parameter", (done)->
+      workWithAuthorPromise
+      .then (work)->
+        { uri:workUri } = work
+        authorUri = work.claims['wdt:P50'][0]
+        nonAuthReq 'get', "#{endpointBase}#{workUri}&relatives=wdt:P50"
+        .then (res)->
+          res.entities[workUri].should.be.an.Object()
+          res.entities[authorUri].should.be.an.Object()
```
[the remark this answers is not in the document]
maxlath: I'm not sure that would be more readable, we would have to define a stable key for each of the [whitelisted relation property](https://github.com/inventaire/inventaire/blob/ff3e3cb/server/controllers/entities/by_uris.coffee#L6-L10) (which isn't much yet but could grow in the future?), while properties URIs already play this role, and you have to be familiar with them to be able to parse the claims anyway

maxlath:
I confirm that it this feature would be useful for the isbn importer to be able to [fetch the works associated to the editions, and those works authors](https://github.com/inventaire/inventaire-client/blob/fcfbf92/app/modules/inventory/lib/import/fetch_entities_sequentially.coffee#L26)

server/controllers/entities/lib/add_relatives.coffee
```diff
@@ -0,0 +1,41 @@
+# Enrich ../by_uris results with entities related to the directly
+# requested entities, following those entities claims
+
+__ = require('config').universalPath
+_ = __.require 'builders', 'utils'
+getEntitiesByUris = require './get_entities_by_uris'
```
jum-s: `add_relatives` file has `get_entities_by_uris` dependency, while it is called right after [get_entities_by_uris](https://github.com/inventaire/inventaire/pull/134/files#diff-e5405d93d05274b0a584fb5fcadd52cbR38). Wouldnt make it easier if `add_relatives` is within `get entities by uris` ?

api_tests/entities/get.test.coffee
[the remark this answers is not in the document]
jum-s: which "properties URIs" do you mean, entity claims ?
i agree changing an array into an object adds complexity. but the readability argument stands, consider this
 
1. `/api/entities?action=by-uris&uris=#{workUri}&relatives=authors`
2. `/api/entities?action=by-uris&uris=#{workUri}&relatives=wdt:P50`

you need to know wikidata DSL to understand point 2. 
making it `{ "authorship": "wdt:P50" }` would be more readable, it is also self-documenting this endpoint. 
and P50 is pretty stable no need to change it in the future.

Providing an API which hides wikidata DSL is a way of making wikidata more popular. The API interface hides DSL understanding, and then when you use this endpoint and would like to know whats behind you dig in inventaires code and for example find out entities use wd claims...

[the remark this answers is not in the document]
maxlath: example of an API consumer code:
```coffee
editionUri = 'isbn:9782330056315'
relatives = [ 'works', 'authors' ]

getEntities { uris: [ editionUri ], relatives }
.then (entities)->
  edition = entities[editionUri]
  # requires to be familiar with the Wikidata properties
  worksUris = edition.claims['wdt:P629']
  works = _.values _.pick(entities, worksUris)
  # requires to be familiar with the Wikidata properties
  authorsUris = _.uniq _.flatten(works.map(_.property('wdt:P50'))
  authors = _.values _.pick(entities, authorsUris)
  # do something with the authors
```

server/controllers/entities/lib/add_relatives.coffee
```diff
@@ -0,0 +1,41 @@
...
+    .then (additionalResults)->
+      # We only need to extend entities, as those additional URIs
+      # should already be the canonical URIs (no redirection needed)
+      # and all URIs should resolve to an existing entity
+      _.extend results.entities, additionalResults.entities
+      return results
+
+  return addRelatives
+
+getAdditionalEntitiesUris = (entities, relatives)->
```
maxlath: @jum-s alternative implementation A:
```coffee
getAdditionalEntitiesUris = (entities, relatives)->
  additionalEntitiesUris = _.values entities
    .map getEntityRelativesUris(relatives)

  return _.uniq _.flattenDeep(additionalEntitiesUris)

getEntityRelativesUris = (relatives)-> (entity)->
  _.values _.pick(entity.claims, relatives)
```

maxlath: @jum-s alternative implementation B:
```coffee
getAdditionalEntitiesUris = (entities, relatives)->
  _(entities)
  .values()
  .map getEntityRelativesUris(relatives)
  .flattenDeep()
  .uniq()
  .value()

getEntityRelativesUris = (relatives)-> (entity)->
  _.values _.pick(entity.claims, relatives)
```

jum-s: very attractive implementation B

maxlath:
re-merged to include 2995ee8 and a0a9a95

PR #137, opened 2018-01-23
Get entity type from wdt:P279

maxlath:
fallback on wdt:P279 claims values for Wikidata entities which type can't be identified from wdt:P31
ex: [Q28135131](https://www.wikidata.org/wiki/Q28135131) currently has a P279 claim that should make it be identified as a genre, but no P31. Consequently, adding it as a genre (wdt:P136) claim value is currently rejected by the editor

### DIFF
--- a/server/controllers/entities/lib/get_entity_type.coffee
+++ b/server/controllers/entities/lib/get_entity_type.coffee
@@ -2,12 +2,21 @@ __ = require('config').universalPath
 _ = __.require 'builders', 'utils'
 { types } =  __.require 'lib', 'wikidata/aliases'
 
-# Takes an entity wdt:P31 claims array
+# Takes an entity wdt:P31 (instance of) claims array
 # Returns a entity type string: work, edition, article, human, genre
-module.exports = (wdtP31Array)->
+# If no type is found, try with the wdt:P279 (subclass of) claims array
+# (used for Wikidata entities only, as all inv entities have a known P31)
+module.exports = (wdtP31Array, wdtP279Array)->
   unless wdtP31Array? then return
 
   for value in wdtP31Array
+    type = types[value]
+    # return as soon as we get a type
+    if type? then return type
+
+  unless wdtP279Array? then return
+
+  for value in wdtP279Array
     type = types[value]
     # return as soon as we get a type
     if type? then return type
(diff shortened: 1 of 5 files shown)


server/controllers/entities/lib/get_entity_type.coffee
```diff
@@ -2,14 +2,23 @@ __ = require('config').universalPath
...
     # return as soon as we get a type
     if type? then return type
 
+  unless wdtP279Array? then return
+
+  for value in wdtP279Array
+    type = types[value]
+    # return as soon as we get a type
+    if type? then return type
+
```
jum-s: why not making it dry for the next entity type to fetch ?

server/controllers/entities/lib/get_wikidata_enriched_entities.coffee
```diff
@@ -45,10 +45,11 @@ getEnrichedEntity = (wdId)->
 
 mergeWdAndInvData = (entity, invEntity)->
   if entity.missing? then return formatEmpty 'missing', entity
-  { P31 } = entity.claims
-  if P31
-    simplifiedP31 = wdk.simplifyPropertyClaims P31
-    entity.type = getEntityType simplifiedP31.map(prefixify)
+  { P31, P279 } = entity.claims
+  if P31? or P279?
```
jum-s: Thats also technical debt, no ? what if 3 other property claims are added next time ?
Cant we use an object or array to store claims and a function to simplify and assign type to entity directly

server/controllers/entities/lib/get_entity_type.coffee
maxlath: how would you write it?
feature to keep: return as soon as we find a matching type

server/controllers/entities/lib/get_wikidata_enriched_entities.coffee
maxlath: [P31 and P279 are the 2 properties explicitly recommended by the Wikidata community to handle typing](https://www.wikidata.org/wiki/Wikidata:Item_classification#Classification_in_Wikidata), and as such requested to be defined, at least one or the other, on each Wikidata entity. Meanwhile, on Inventaire, each entity has a P31 claim and no P279

maxlath: [P31 and P279 are the 2 properties explicitly recommended by the Wikidata community to handle typing](https://www.wikidata.org/wiki/Wikidata:Item_classification#Classification_in_Wikidata), and as such requested to be defined, at least one or the other, on each Wikidata entity. Meanwhile, on Inventaire, each entity has a P31 claim and no P279

server/controllers/entities/lib/get_entity_type.coffee
maxlath: how would you write it?
feature to keep: return as soon as we find a matching type

PR #138, opened 2018-01-23
entities: reverse claims: blacklisting properties with too many results

maxlath:
namely wdt:P31
Motivations:
- no use case for knowing all the entities matching a given P31
- allowing to query all the instances of Q571 or Q5 sorted by popularity was making the server eat all the RAM


jum-s:
oh and +1 for the tests :)

maxlath:
none of P50, P18, and P144 would currently return more than 1000 results (ex: in the case 1000 works are linking to an author with P50, which seems quite extreme), while the P31:Q5 reverse claim would return 250000 results in inv entities alone

PR #146, opened 2018-02-22
cleaning properties

maxlath:
properties configuration object were a bit hard to read for new comers:
- splitted and restructed the main file
- removed the unnecessary `format: _.identity` functions
- took the occasion of cleaning a property config validation test to do some cleaning on string validation functions

### DIFF
--- a/api_tests/entities/update_claims.test.coffee
+++ b/api_tests/entities/update_claims.test.coffee
@@ -23,6 +23,23 @@ describe 'entities:update-claims', ->
 
     return
 
+  it 'should reject an update with an inappropriate property datatype', (done)->
+    createWork()
+    .then (work)->
+      authReq 'put', '/api/entities?action=update-claim',
+        id: work._id
+        # A work entity should not have pages count
+        property: 'wdt:P50'
+        'new-value': 124
+      .then undesiredRes(done)
+      .catch (err)->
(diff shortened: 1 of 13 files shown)


maxlath:
I stand by the datatype and primordial datatype distinction for the purpose of better error report: if the API expects a `positive-integer` and I send a string number value (ex: `"123"`), without primordial type check, the error would be `invalid positive-integer`, giving no clue that the issue is that I passed a string instead of a number.

jum-s:
commit `7498dc1` is trying to improve test coverage to cover the case we are discussing about

do you agree with this test, i feel Im missing something here since test fails on [this line](https://github.com/inventaire/inventaire/blob/365e726e29eb0b2778a1e71db81972bd38466681/server/controllers/entities/lib/validate_claim_value.coffee#L31) instead of failing [right after](https://github.com/inventaire/inventaire/blob/365e726e29eb0b2778a1e71db81972bd38466681/server/controllers/entities/lib/validate_claim_value.coffee#L34)
is there any better mock to trigger than [this](https://github.com/inventaire/inventaire/blob/7498dc1cb1010bbe3e9cb4dbf88d8ecc80847843/api_tests/entities/update_claims.test.coffee#L33) ?

maxlath:
now with 01b0c22, the test pass

PR #150, opened 2018-03-01
Syntax lint rulez

jum-s:
make use of warnings in coffeelint

for `cyclomatic_complexity` and `unnecessary_double_quotes`


maxlath:
nice! the only problem is that I don't know how to resolve those `cyclomatic_complexity` warnings :/

Also, could you rebase on the current master? You will have a conflict with `server/data/wikipedia/extract.coffee` due to f3462cf, but that shouldn't be too hard to fix

maxlath:
re-merged as 7b633df to keep history clean

PR #151, opened 2018-03-01
patches byUserId: refactored to let CouchDB do the pagination work

maxlath:
instead of doing it in the server, which was highly inefficient for large sets of rows
Fetching contributions from users with 10,000+ contributions was triggering timeouts,
now it's done within 50ms

[This PR is given a "low priority" label has this only affects admins]


maxlath:
re-merged in b1c54a3 to keep history clean

PR #153, opened 2018-03-02
entities: fix update_inv_claim for obsolete entities

maxlath:
it was throwing `500` errors when requested to update obsolete entities


api_tests/entities/update_claims.test.coffee
```diff
@@ -23,7 +24,7 @@ describe 'entities:update-claims', ->
 
     return
 
-  it 'should reject an update removing a critical claim', (done)->
+  xit 'should reject an update removing a critical claim', (done)->
```
jum-s: oups! also on L10

maxlath: commit amended: befa921

PR #154, opened 2018-03-03
Rename tests to validations

jum-s:
massive renaming + little gardening
all tests (unit + integration) pass 
so we'll see if it breaks in production in order to improve test coverage :)


server/controllers/items/by_user_and_entity.coffee
```diff
@@ -28,7 +28,7 @@ getUser = (query, reqUserId)->
     return user_.getUserById userId, reqUserId
 
   else if username?
-    unless tests.username username
+    unless isValid username
```
maxlath: this should crash: `Username` and thus `isValid` is a regex, not a function.
Being function is the role of `__.require('models', 'validations/user').username`, so I would prefer something like this:
```coffee
{ username: isValidUsername } = __.require 'models', 'validations/user'
[...]
unless isValidUsername username
```

server/controllers/auth/reset_password.coffee
```diff
@@ -3,13 +3,13 @@ __ = require('config').universalPath
...
 user_ = __.require 'controllers', 'user/lib/user'
-testEmail = __.require('models', 'tests/user').email
+validate = __.require('models', 'validations/user').email
 # pw_ = __.require('lib', 'crypto').passwords
 
 module.exports = (req, res, next)->
   { email } = req.body
   unless email? then return error_.bundleMissingBody req, res, 'email'
-  unless testEmail email
+  unless validate email
```
maxlath: can't the function keep a more precise name like `validateEmail`?

jum-s: `isValid`?

PR #158, opened 2018-03-18
101 boost entities popularity

jum-s:
Boost entity globalScore according to the popularity score and the elastic lexical score.
Should be good to try out !


maxlath:
[meta] there was an error in the `Co-authored-by` mentions: it should be `Co-authored-by: maxlath <k@maxlath.eu>`: do you mind amending the commits?

Issue #160, opened 2018-03-20
lower popularity calculation weight on resources


maxlath:
now that #101 was fixed by #158, approximately each key stroke in the live search input triggers a call on `/api/search`, which request the popularity of the returned entities, which often results in adding new entities to the queue of entities needing their popularity to be calculated: this results in hundreds of entities being added to the queue, which does affect the server's performances. Having a huge queue also results in a larger RAM consumption and did crash the prod in the past already.

A solution could be to persist the queue in LevelDB instead of the RAM, lowering the queue pace, and, in prod, fully delegating the calculation to the alt instance (an instance dedicated to background tasks, like activity summaries, etc)

maxlath:
fixed by #161

PR #161, opened 2018-03-21
160 lower popularity calculation weight on resources

maxlath:
addressing #160


maxlath:
added commits b56b6e1, f2559da, and 49f9625 and remerged as d9b29c7

Issue #163, opened 2018-03-27
Search results that match partial string search


jum-s:
When search query is equal to "Jules ver", results display "Robert Ver" before "Jules Verne" 
Its happening because result match only on [spaced words](https://github.com/inventaire/inventaire/blob/52e405df4b43dfce9581a4e4dcfddecdc42619c4/server/controllers/entities/lib/common_query_body_builder.coffee#L7), instead of full query match.
It might be fixed by changing the "phrase_prefix" [term](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-multi-match-query.html#type-phrase)

maxlath:
the same request as in the comment above returns no result at all at the moment

jum-s:
[Recent](https://github.com/inventaire/inventaire/pull/462) [changes](https://github.com/inventaire/inventaire/pull/478) [improved](https://github.com/inventaire/inventaire/pull/165) partial matches (as there is no `phrase_prefix` anymore). 
"Robert Ver" is not in the result anymore when searching "Jules Ver". And "Jules Vern" does return Q33977 (among Jules Vernon and others but that is not what this issue was reporting).
There is still space for improving partial match in general, but this may be done in a new issue

PR #165, opened 2018-03-28
improve elastic search query

jum-s:
remove last term boost
add match_phrase and match_phrase_prefix
to boost the whole query instead of individual terms

### DIFF
--- a/server/controllers/search/lib/query_body_builder.coffee
+++ b/server/controllers/search/lib/query_body_builder.coffee
@@ -4,6 +4,7 @@ _ = __.require 'builders', 'utils'
 
 module.exports = (search, size)->
   should = [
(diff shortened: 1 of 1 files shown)


server/controllers/search/lib/query_body_builder.coffee
```diff
@@ -4,6 +4,7 @@ _ = __.require 'builders', 'utils'
 
 module.exports = (search, size)->
   should = [
+    { match_phrase_prefix: { _all: { query: search, boost: 5 } } }
     { match: { _all: { query: search, boost: 5 } } }
     { prefix: { _all: _.last(search.split(' ')) } }
```
maxlath: so we keep `{ prefix: { _all: _.last(search.split(' ')) } }` finally?

maxlath:
addressing #163

PR #167, opened 2018-04-12
restructure config file to run jobs indepentently

jum-s:


server/db/level/jobs.coffee
```diff
@@ -8,9 +8,9 @@ module.exports =
   # taking two arguments; a payload and an error logger function
   initQueue: (jobName, worker, maxConcurrency)->
     db = levelBase.sub "job:#{jobName}"
-
-    # If runJobsInQueue is true, push and run jobs
-    if CONFIG.runJobsInQueue
+    # Only starts job marked in runJobsInQueue, push and run jobs
+    jobKey = jobName.split(':')[1]
```
maxlath: using the full `jobName` as key seems less hazardous, as it let's more freedom in the `jobName` definition

server/db/level/jobs.coffee
```diff
@@ -8,9 +8,9 @@ module.exports =
   # taking two arguments; a payload and an error logger function
   initQueue: (jobName, worker, maxConcurrency)->
     db = levelBase.sub "job:#{jobName}"
-
-    # If runJobsInQueue is true, push and run jobs
-    if CONFIG.runJobsInQueue
+    # Only starts job marked in runJobsInQueue, push and run jobs
```
maxlath: proposition of rephrasing: `# If the job is whitelisted, push and run jobs`

server/db/level/jobs.coffee
jum-s: its prefixyied (jobname  = `wd:popularity`) 
do we need a key without a `:` in the middle for config files ?

server/db/level/jobs.coffee
jum-s: is it a whtelist tho, its a key value configuration not a hardcoded stored array.
maybe `flagged` instead of `marked`

server/db/level/jobs.coffee
maxlath: no, it's ok to have a `:` in the key, it just requires to use single quotes around the key:

:negative_squared_cross_mark: crashes:
```coffee
runJobsInQueue:
  wd:popularity: true
```
:white_check_mark: works:
```coffee
runJobsInQueue:
  'wd:popularity': true
```

server/db/level/jobs.coffee
maxlath: maybe `flagged`, but the beginning of the phrase still doesn't make sense with the `push and run jobs` part, no?

PR #170, opened 2018-04-22
Adding entities tests

maxlath:
this is a bulk of atomic cleaning commits on entities, see commit messages for details


api_tests/entities/update_labels.test.coffee
```diff
@@ -32,3 +32,18 @@ describe 'entities:update-labels', ->
     .catch undesiredErr(done)
 
     return
+
+  it 'should reject an up-to-date value', (done)->
+    humanPromise
+    .then (human)->
+      updateLabel human._id, 'en', 'foo'
+      .catch undesiredErr(done)
+      .then -> updateLabel human._id, 'en', 'foo'
```
jum-s: suggestion : 
```
    humanPromise
    .then (human)->
      humanLabel = human.labels.en
      updateLabel human._id, 'en', humanLabel
```

server/controllers/entities/lib/entities.coffee
```diff
@@ -97,6 +97,8 @@ module.exports = entities_ =
   putUpdate: (params)->
     { userId, currentDoc, updatedDoc } = params
     _.types [ userId, currentDoc, updatedDoc ], ['string', 'object', 'object']
+    # Checking that there is actually an update is let to the responsability
+    # of consumers: empty patches at this stage will throw 500 errors
```
jum-s: suggestion : 

    # It is to the consumers responsability to check if there is an update: 
    # empty patches at this stage will throw 500 errors

server/controllers/entities/lib/get_inv_entities.coffee
```diff
@@ -10,7 +12,10 @@ module.exports = (ids, refresh)->
   # as inv entities with an associated Wikidata entity use the Wikidata uri
   entities_.byIds ids
   .map Format(refresh)
-  .then (entities)-> { entities }
+  .then (entities)->
+    found = _.pluck entities, '_id'
+    notFound = _.difference(ids, found).map prefixify
```
jum-s: merci lodash !

api_tests/entities/update_labels.test.coffee
maxlath: `'foo'` represents a new value, not the old one: the old one isn't needed for labels update (at the difference of claims) as there is only one value per language: it could benefit from being renamed `set-label` instead of `update-label`

PR #173, opened 2018-05-07
Populate tests fixtures

jum-s:
Scripts that POST `/_replicate` on couchdb 
from `*-tests` databases to `*`
an unobstrusive default replicate behavior

### DIFF
--- a/scripts/populate_users_fixtures.coffee
+++ b/scripts/populate_users_fixtures.coffee
@@ -1,0 +1,14 @@
+#!/usr/bin/env coffee
+CONFIG = require 'config'
+__ = CONFIG.universalPath
+_ = __.require 'builders', 'utils'
+{ createUserWithItems } = require '../api_tests/fixtures/populate'
+
+createUserWithItems()
+.then (userCreated)->
+  _.success '#### New User available ####'
+  console.log "Your can now login with :\n\
+    - Username : #{userCreated.username} \n\
+    - Password : #{userCreated.username}"
+  process.exit 0
+.catch (err)-> console.log err
--- a/scripts/populate_users_fixtures.coffee
+++ b/scripts/populate_users_fixtures.coffee
@@ -9,6 +9,6 @@ createUserWithItems()
   _.success '#### New User available ####'
   console.log "Your can now login with :\n\
     - Username : #{userCreated.username} \n\
-    - Password : #{userCreated.username}"
+    - Password : 12345678"
   process.exit 0
-.catch (err)-> console.log err
+.catch _.Error err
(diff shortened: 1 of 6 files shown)


scripts/replicate_tests_db.coffee
```diff
@@ -0,0 +1,24 @@
...
+{ exec } = require 'child_process'
+promises_ = __.require 'lib', 'promises'
+
+dbHost = require('config').db.fullHost()
+allDbsUrl = dbHost + '/_all_dbs'
+
+tests_dbs = (db)-> db.includes '-tests'
+dbUrl = (dbName)-> "#{dbHost}/#{dbName}"
+
+promises_.get allDbsUrl
```
maxlath: you actually don't need to fetch those dbs as you already have the list at hand:
```coffee
dbsBaseNames = Object.keys __.require('couch', 'list')
```

jum-s: this [list](inventaire/server/db/couch/list.coffee) do not contain `-tests` dbs. I would rather trust an explicit response from couch than a maintained hardcoded list.

scripts/populate_users_fixtures.coffee
```diff
@@ -0,0 +1,14 @@
...
+__ = CONFIG.universalPath
+_ = __.require 'builders', 'utils'
+{ createUserWithItems } = require '../api_tests/fixtures/populate'
+
+createUserWithItems()
+.then (userCreated)->
+  _.success '#### New User available ####'
+  console.log "Your can now login with :\n\
+    - Username : #{userCreated.username} \n\
+    - Password : #{userCreated.username}"
```
maxlath: now that ed2b027 was merged on master, this needs to be updated

scripts/populate_users_fixtures.coffee
```diff
@@ -0,0 +1,14 @@
...
+  process.exit 0
+.catch (err)-> console.log err
```
maxlath: this could take advantage of the "standard" way to log errors: `.catch _.Error('some label')`

scripts/replicate_tests_db.coffee
```diff
@@ -0,0 +1,29 @@
...
+  dbTestName = "#{dbName}-tests"
+  repDoc =
+    source: dbUrl dbTestName
+    target: dbUrl dbName
+  breq.post "#{dbHost}/_replicate", repDoc
+  .then (res)-> res.body
+  .then _.Log "#{dbTestName} replication response"
+.catch (err)->
+  console.log "#{err.body.reason}\n\
+  Hum, have you ran the tests first ?"
```
maxlath: shouldn't `scripts/populate_fixtures` make sure tests are run instead?

jum-s: this script replicate dbs, its the user responsability to replicate dbs that exist already. It is written in the documentation where [this](https://github.com/inventaire/docker-inventaire/blob/master/README.md#fixtures) script is described

maxlath:
@jum-s I added a linting commit d569628, now ready to rebase and merge if that's good for you

PR #175, opened 2018-05-09
level base: cleaning

maxlath:
changing the `server/db/level/base` interface to only keep what we actually need: a way to get a raw sublevel db (`rawSubDb`) to be used in combinations with level db modules (namely `level-jobs` and `level-geospatial`), and a simplified (=promisified and shipping with some helper functions) db `simplifiedSubDb`


server/db/level/base.coffee
```diff
@@ -15,51 +15,46 @@ else
...
-    sub = @sub dbName
-    API = @unjsonized @promisified(sub)
-    API.sub = sub
-    return API
+rawSubDb = (dbName, valueEncoding)->
+  _.success "#{dbName} opened"
+  return DB.sublevel dbName, { valueEncoding }
+
+# Promisified and with a few additional functions
+simplifiedSubDb = (dbName)->
```
jum-s: simplified ? looks more like 'simpleSubDb' no?
simplified implies simplified compare to something.

maxlath: yes, compared to the `rawSubDb`

PR #177, opened 2018-05-10
Move items entity snapshot to leveldb

maxlath:
based on: #175 (@jum-s please merge review and merge #175 first)
main commit: 283905a5

Taking snapshot on-the-fly in a LevelDB database instead of having a copy in each item CouchDB document, so that items with the same entity mutualize that cache

Benefits:
  - keep item docs clean of non-master data
  - removes the need to update batches of documents in the database on every entity change,
    it's instead done onces in the snapshot leveldb
Inconvenient:
  - having several snapshot for work entities used as entity by items (instead of editions)
    was creating too much complexity (having the keep the lang on the item doc + a flag to tell it was a work entity)
    so only one snapshot per work is available instead, using the work original languages.
    While creating a possible annoyance for users seeing data in the wrong language, it creates an insentive
    to precise the edition.


server/controllers/items/lib/snapshot/refresh_snapshot.coffee
```diff
@@ -1,28 +1,28 @@
...
 entities_ = __.require 'controllers', 'entities/lib/entities'
 getEntityByUri = __.require 'controllers', 'entities/lib/get_entity_by_uri'
 getInvEntityCanonicalUri = __.require 'controllers', 'entities/lib/get_inv_entity_canonical_uri'
 buildSnapshot = require './build_snapshot'
 { getWorkAuthorsAndSeries, getEditionGraphEntities } = require './get_entities'
 { getDocData } = require './helpers'
-items_ = require '../items'
-Item = __.require 'models', 'item'
+
+# Working around circular dependencies
```
jum-s: why not deleting it instead :) ?

test/models/023-item.coffee
```diff
@@ -140,17 +123,3 @@ describe 'item model', ->
...
-      updateAttributesData =
-        snapshot:
-          'entity:image': '/ipfs/QmXtPA8rAn5VCJ3rVLxWhDEnMv2hiPTMopH9NuaERz3nuw'
-      (-> update(updateAttributesData, doc)).should.not.throw()
-      done()
-
-    it 'should throw when updated with an invalid snapshot', (done)->
-      doc = create validItem
-      updateAttributesData = { snapshot: { bla: 'hello' } }
-      (-> update(updateAttributesData, doc)).should.throw()
```
jum-s: is this deletion tested somewhere else now. maybe in `/api_tests/items/snapshot.test.coffee` ?

jum-s: is it fine without testing ? :blush:

maxlath: now that the snapshot aren't written in items documents anymore, the closest possible would be to test the schema of snapshot data attached to items, [which is done here](https://github.com/inventaire/inventaire/blob/29fd78c/tests/api/items/snapshot.test.coffee). so I think it's fine without additional test

PR #178, opened 2018-05-10
entities merge: fixed merge from an entity with an ISBN

maxlath:
make the [`should merge an entity with an ISBN` test](https://github.com/inventaire/inventaire/blob/70684f6/api_tests/entities/merge.test.coffee#L53) pass


maxlath:
(not really voluntarily) re-done on the `fix-tests` branch (I knew I had already been there ><), see #193 193

PR #194, opened 2018-06-15
Duplicate shared libs

maxlath:
to work around out-of-sync branches errors by tracking shared libs on each repos
    
We keep it DRY by replicating client shared lib folder in the server own folder when branches are in sync, using the post-checkout git hook

### DIFF
--- a/.githooks/post-checkout/post-checkout
+++ b/.githooks/post-checkout/post-checkout
@@ -1,0 +1,2 @@
+#!/usr/bin/env zsh
+./scripts/sync_shared_libs
--- a/.githooks/post-checkout/post-checkout
+++ b/.githooks/post-checkout/post-checkout
@@ -1,2 +1,0 @@
-#!/usr/bin/env zsh
-./scripts/sync_shared_libs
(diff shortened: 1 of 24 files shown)


jum-s:
a third party lib repo would make sync easily adaptable whatever working on server or client side. It also offers the possibility to develop a new independent client without being dependent of shared libs stored in the current client. 
Small independent libs is a nice way to think about all those common dependecies

maxlath:
that was the goal of [`inv-utils`](https://github.com/inventaire/inv-utils), [`inv-couch`](https://github.com/inventaire/inv-couch), [`inv-loggers`](https://github.com/inventaire/inv-loggers), or even [`inv-dbs-list`](https://github.com/inventaire/inv-dbs-list), but that didn't really work-out: every minor change meant having to release a new version on npm and update the package.json ([examples](https://github.com/inventaire/inventaire/search?q=inv-utils&type=Commits)). I would prefer copying and pasting code rather than getting back to this workflow ;)

maxlath:
made obsolete by #237

PR #199, opened 2018-06-26
tasks: filter not already suspect entities with refresh flag

jum-s:


maxlath:
@jum-s is this PR obsolete or should it be rebased+conflict resolved and merged?

jum-s:
not obsolete, would still be nice to not check every entities all over at each collect entities

maxlath:
@jum-s I added de59072a d2ab5d11 and 3dd4d6f2, can you review those?

PR #222, opened 2018-10-24
Tasks reconciler bot

jum-s:
Merge entities if no homoymns suggestion and first work label longer than 12
If no possibilities of merge then create a task with a detailed occruences as
task.hasEncyclopediaOccurence: 
```
  [
    { 'uri': 'wd:xxx',
    'url': 'http://en.wikipedia.org/xxx'
    'matchedPatten': '/worklabel/gi' }
  ]
```

### DIFF
--- a/db/couchdb/design_docs/tasks.json
+++ b/db/couchdb/design_docs/tasks.json
@@ -7,6 +7,9 @@
     },
     "byScore": {
       "map": "(doc)->\n  unless doc.state \n    emit [doc.hasEncyclopediaOccurence, doc.relationScore, doc.elasticScore], doc"
+    },
+    "hasEncyclopediaOccurence": {
+      "map": "(doc)->\n  unless doc.state \n    if doc.hasEncyclopediaOccurence == true\n      emit doc.id, null"
     }
   }
 }
--- a/server/controllers/entities/lib/has_works_labels_occurrence.coffee
+++ b/server/controllers/entities/lib/has_works_labels_occurrence.coffee
@@ -23,7 +23,7 @@ module.exports = (wdAuthorUri, worksLabels, worksLabelsLangs)->
   # Filter-out labels that are too short, as it could generate false positives
   worksLabels = worksLabels.filter (label)-> label.length > 5
 
-  if worksLabels.length is 0 then return promises_.resolve false
+  if worksLabels.length is 0 then return promises_.resolve []
 
   # get Wikipedia article title from URI
   getEntityByUri wdAuthorUri
@@ -34,7 +34,7 @@ module.exports = (wdAuthorUri, worksLabels, worksLabelsLangs)->
       hasWikipediaOccurrence authorEntity, worksLabels, worksLabelsLangs
       hasBnfOccurrence authorEntity, worksLabels
     ]
-  .spread (wpBool, bnfBool)-> wpBool or bnfBool
+  .then (res)-> _.compact(_.flatten(res))
   .catch (err)->
     _.error err, 'has works labels occurrence err'
     # Default to false if an error happened
@@ -45,14 +45,12 @@ hasWikipediaOccurrence = (authorEntity, worksLabels, worksLabelsLangs)->
   .then (articles)->
     # Match any of the works labels
     worksLabelsPattern = new RegExp(worksLabels.join('|'), 'gi')
-    for article in articles
-      if article.extract.match(worksLabelsPattern)? then return true
-    return false
+    articles.map createOccurenceDocs worksLabelsPattern, authorEntity
 
 getMostRelevantWikipediaArticles = (authorEntity, worksLabelsLangs)->
   { sitelinks, originalLang } = authorEntity
 
-  return _.uniq worksLabelsLangs.concat([ originalLang, 'en' ])
+  return _.uniq worksLabelsLangs.concat [ originalLang, 'en' ]
   .map (lang)->
     title = sitelinks["#{lang}wiki"]
     if title? then return { lang, title }
@@ -69,4 +67,13 @@ hasBnfOccurrence = (authorEntity, worksLabels)->
   # is wrong and we can't know which
   if bnfIds?.length isnt 1 then return false
   getBnfAuthorWorksTitles bnfIds[0]
-  .then (titles)-> _.haveAMatch titles, worksLabels
+  .then createOccurencDocs worksLabelsPattern, authorEntity
+
+createOccurenceDocs = (worksLabelsPattern, authorEntity)->
+  return (article)->
+    unless article.extract.match(worksLabelsPattern)?
+      return []
+    return
+      uri: authorEntity.uri
+      url: article.url
+      patternMatch: worksLabelsPattern
--- a/server/controllers/tasks/lib/build_task_docs.coffee
+++ b/server/controllers/tasks/lib/build_task_docs.coffee
@@ -2,9 +2,9 @@ __ = require('config').universalPath
 _ = __.require 'builders', 'utils'
 { Promise } = __.require 'lib', 'promises'
 
-entities_ = __.require 'controllers', 'entities/lib/entities'
 searchEntityDuplicatesSuggestions = require './search_entity_duplicates_suggestions'
 { calculateRelationScore } = require './relation_score'
+getAuthorWorksData = require './get_author_works_data'
 hasWorksLabelsOccurrence = __.require 'controllers', 'entities/lib/has_works_labels_occurrence'
 { turnIntoRedirection } = __.require 'controllers', 'entities/lib/merge_entities'
 { prefixifyInv } = __.require 'controllers', 'entities/lib/prefix'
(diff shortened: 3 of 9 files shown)


db/couchdb/design_docs/tasks.json
```diff
@@ -7,6 +7,9 @@
     },
     "byScore": {
       "map": "(doc)->\n  unless doc.state \n    emit [doc.hasEncyclopediaOccurence, doc.relationScore, doc.elasticScore], doc"
+    },
+    "hasEncyclopediaOccurence": {
```
maxlath: where do you call this view?

server/controllers/entities/lib/has_works_labels_occurrence.coffee
```diff
@@ -45,14 +45,12 @@ hasWikipediaOccurrence = (authorEntity, worksLabels, worksLabelsLangs)->
   .then (articles)->
     # Match any of the works labels
     worksLabelsPattern = new RegExp(worksLabels.join('|'), 'gi')
-    for article in articles
-      if article.extract.match(worksLabelsPattern)? then return true
-    return false
+    articles.map createOccurenceDocs worksLabelsPattern, authorEntity
```
maxlath:
```suggestion
    articles.map createOccurenceDocs(worksLabelsPattern, authorEntity)
```

server/controllers/entities/lib/has_works_labels_occurrence.coffee
```diff
@@ -45,14 +45,12 @@ hasWikipediaOccurrence = (authorEntity, worksLabels, worksLabelsLangs)->
...
 
 getMostRelevantWikipediaArticles = (authorEntity, worksLabelsLangs)->
   { sitelinks, originalLang } = authorEntity
 
-  return _.uniq worksLabelsLangs.concat([ originalLang, 'en' ])
+  return _.uniq worksLabelsLangs.concat [ originalLang, 'en' ]
```
maxlath:
```suggestion
  return _.uniq worksLabelsLangs.concat([ originalLang, 'en' ])
```

server/controllers/entities/lib/has_works_labels_occurrence.coffee
```diff
@@ -69,4 +67,13 @@ hasBnfOccurrence = (authorEntity, worksLabels)->
   # is wrong and we can't know which
   if bnfIds?.length isnt 1 then return false
   getBnfAuthorWorksTitles bnfIds[0]
-  .then (titles)-> _.haveAMatch titles, worksLabels
+  .then createOccurencDocs worksLabelsPattern, authorEntity
```
maxlath:
```suggestion
  .then createOccurencDocs(worksLabelsPattern, authorEntity)
```

server/controllers/tasks/lib/build_task_docs.coffee
```diff
@@ -2,49 +2,82 @@ __ = require('config').universalPath
...
-
-create = (authorWorksData, relationScore)->
-  { labels, langs } = authorWorksData
-  return (suggestionEntity)->
-    suggestionUri = suggestionEntity.uri
-    _.type suggestionUri, 'string'
-    hasWorksLabelsOccurrence suggestionUri, labels, langs
-    .then (hasOccurence)->
+  .spread (suggestions, authorWorksData)->
+    unless _.some suggestions then return []
```
maxlath:
```suggestion
    unless suggestions.length > 0 then return []
```
no need to call [such a sophisticated function](https://github.com/lodash/lodash/blob/b97a9a3da01077337540a377bf87860c4d00db2c/collection/some.js#L56-L65) for such a simple test, no?

server/controllers/tasks/lib/build_task_docs.coffee
```diff
@@ -2,49 +2,82 @@ __ = require('config').universalPath
...
+    Promise.all suggestions.map getOccurences authorWorksData
```
maxlath:
```suggestion
    Promise.all suggestions.map(getOccurences(authorWorksData))
```

server/controllers/tasks/lib/build_task_docs.coffee
```diff
@@ -2,49 +2,82 @@ __ = require('config').universalPath
...
+    .then turnSuggestionIntoRedirection(suggestions, authorWorksData)
+    .then (occurences)->
+      Promise.all filterSuggestions occurences, suggestions
```
maxlath:
```suggestion
      Promise.all filterSuggestions(occurences, suggestions)
```

server/controllers/tasks/lib/build_task_docs.coffee
```diff
@@ -2,49 +2,82 @@ __ = require('config').universalPath
...
+    .then (occurences)->
+      if _.isEmpty(_.flattenDeep(occurences)) then return []
+      return
+        uri: suggestion.uri
+        occurences: occurences
+
+create = (authorWorksData, relationScore, suggestionsOccurences)->
+  return (suggestion)->
+    { authorId } = authorWorksData
+    suggestionOccurences = _.filter(suggestionsOccurences, {'uri': suggestion.uri})
```
maxlath: this means looping through `suggestionsOccurences.length` times over.
A work around would be to create an index in the higher order function:
```coffee
# between L70-L71
occurrencesBySuggestionUri = _.indexBy suggestionsOccurences, 'uri'
```
```coffee
suggestionOccurences = occurrencesBySuggestionUri[suggestion.uri]
```
(unless there are several `suggestionsOccurences` with the same uri, in which case, why taking only the first one into account here after?)

server/controllers/tasks/lib/build_task_docs.coffee
```diff
@@ -2,49 +2,82 @@ __ = require('config').universalPath
...
-    worksData = works.reduce aggregateWorksData, base
-    worksData.langs = _.uniq worksData.langs
-    return worksData
-
-aggregateWorksData = (worksData, work)->
-  for lang, label of work.labels
-    worksData.labels.push label
-    worksData.langs.push lang
-  return worksData
+      hasEncyclopediaOccurence: occurences or {}
```
maxlath: `hasEncyclopediaOccurence` was hinting to a boolean value, this could be renamed something like `encyclopediaOccurences`, or, for non-encyclopedic future-proofness, something like `externalSourcesOccurrences`

maxlath: `occurences`, if defined, is an array, so why default to an object?

server/controllers/entities/lib/has_works_labels_occurrence.coffee
```diff
@@ -69,4 +67,13 @@ hasBnfOccurrence = (authorEntity, worksLabels)->
...
+
+createOccurenceDocs = (worksLabelsPattern, authorEntity)->
```
maxlath: in the rest of the code, `docs` is associated with full CouchDB documents: what about `createOccurenceObj`?

tests/api/tasks/has_encyclopedia_occurence.test.coffee
```diff
@@ -5,19 +5,90 @@ should = require 'should'
...
+    humanLabel = 'Stanislas Lem' # has no homonyms
+    workLabel = 'Solaris' # too short label to be automerged
+    createHuman { labels: { en: humanLabel } }
+    .then (human)->
+      createWorkWithAuthor human, workLabel
+      .then (work)-> checkEntities human.uri
+      .then (tasks)->
+        tasks.length.should.aboveOrEqual 1
+        task = tasks.find (task)-> task.suggestionUri.match /wd:/
+        task.hasEncyclopediaOccurence[0].uri.should.exist
```
maxlath: `task.hasEncyclopediaOccurence[0].uri.should.exist` is `undefined`
```suggestion
        task.hasEncyclopediaOccurence[0].uri.should.be.ok()
```

server/controllers/entities/lib/has_works_labels_occurrence.coffee
```diff
@@ -69,4 +67,13 @@ hasBnfOccurrence = (authorEntity, worksLabels)->
...
+  return (article)->
+    unless article.extract.match(worksLabelsPattern)?
+      return []
+    return
+      uri: authorEntity.uri
+      url: article.url
+      patternMatch: worksLabelsPattern
```
maxlath: regex can't be saved as such as they aren't part of the JSON standard:
```coffee
JSON.stringify { foo: /bar/ }
# => '{"foo":{}}'
```
But anyway, storing the full `worksLabelsPattern` isn't that useful, the really interesting info would be to know which title matched

server/controllers/entities/lib/has_works_labels_occurrence.coffee
jum-s: or `createOccurences`

tests/api/tasks/has_encyclopedia_occurence.test.coffee
jum-s: hum, worked for me, but changed anyway

maxlath: the test does pass, it just doesn't test anything

server/controllers/entities/lib/has_works_labels_occurrence.coffee
```diff
@@ -63,10 +59,19 @@ getWikipediaArticleFromSitelinkData = (sitelinkData)->
...
   # is wrong and we can't know which
   if bnfIds?.length isnt 1 then return false
   getBnfAuthorWorksTitles bnfIds[0]
-  .then (titles)-> _.haveAMatch titles, worksLabels
+  .map createOccurrences(worksLabels)
+
+createOccurrences = (worksLabels)->
+  worksLabelsPattern = new RegExp(worksLabels.join('|'), 'gi')
+  return (article)->
+    matchedTitles = _.uniq article.quotation.match(worksLabelsPattern)
```
maxlath: nice!

server/controllers/tasks/lib/build_task_docs.coffee
```diff
@@ -2,49 +2,86 @@ __ = require('config').universalPath
...
+
+createTasksDocs = (authorWorksData, occurrences) ->
+  return (suggestions) ->
+    relationScore = calculateRelationScore suggestions
+    return suggestions.map create(authorWorksData, relationScore, occurrences)
+
+canBeRedirected = (suggestions, matchedTitles) ->
+  # several suggestions == has homonym => cannot be redirected
+  unless suggestions.length == 1 then return false
+  longTitles = matchedTitles.filter (title) -> title.length > 12
```
maxlath: shouldn't there be also a critiria in the number of words?
```suggestion
  longTitles = matchedTitles.filter (title) -> title.length > 12 and title.split(' ').length > 1
```

jum-s: not sure its necessary, a word of 12 chars is as radom as 2 words of 12 chars total, which is also as random as 6 words of 1é chars total

maxlath: I don't think this statement is true, but we can still try like that and see what kind and quantity of false positive it generates, and adapt accordingly

PR #231, opened 2018-11-06
Fix items snapshot image

maxlath:
well, that was ugly

the issue was that [non-formatted edition docs (that is, without .image.url attribute) were used when refreshing snapshot following an event on the work](https://github.com/inventaire/inventaire/blob/a9fecad/server/controllers/items/lib/snapshot/refresh_snapshot.coffee#L67), creating this weird situation were snapshot refresh triggered by the edition returned with images but not when triggered from the work...


jum-s:
vive le test automatisé !

PR #232, opened 2018-11-07
Fix user delete

maxlath:
cleaning to prevent a 500 error because of a [missing `comments_` variable](https://github.com/inventaire/inventaire/blob/bf45de0/server/controllers/comments/lib/comments.coffee)

### DIFF
--- a/server/controllers/items/lib/delete_user_items.coffee
+++ b/server/controllers/items/lib/delete_user_items.coffee
@@ -6,16 +6,5 @@ items_ = __.require 'controllers', 'items/lib/items'
 promises_ = __.require 'lib', 'promises'
 
 module.exports = (userId)->
-  # get user items ids
   items_.byOwner userId
-  .then deleteItemsAndComments.bind(null, userId)
-
-deleteItemsAndComments = (userId, items)->
-  promises_.all [
-    items_.bulkDelete items
-    deleteCommentsByItems userId, items
-  ]
-
-deleteCommentsByItems = (userId, items)->
-  itemsIds = items.map _.property('_id')
-  comments_.deleteByItemsIds itemsIds
+  .then items_.bulkDelete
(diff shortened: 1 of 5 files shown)


tests/api/user/delete.test.coffee
```diff
@@ -0,0 +1,32 @@
...
+    .then (user)->
+      customAuthReq userPromise, 'delete', '/api/user'
+      .then (res)->
+        res.ok.should.be.true()
+        getRefreshedUser userPromise
+        .then (deletedUser)->
+          deletedUser._id.should.equal user._id
+          previousRevInteger = parseInt(user._rev .split('-')[0])
+          parseInt(deletedUser._rev.split('-')[0]).should.equal(previousRevInteger + 1)
+          deletedUser.username.should.equal user.username
```
jum-s: keeping username slot occupied ? what about i delete my account and someoneelse want to user de username i used

server/controllers/items/lib/delete_user_items.coffee
```diff
@@ -6,16 +6,5 @@ items_ = __.require 'controllers', 'items/lib/items'
 promises_ = __.require 'lib', 'promises'
 
 module.exports = (userId)->
-  # get user items ids
   items_.byOwner userId
-  .then deleteItemsAndComments.bind(null, userId)
```
jum-s: what about delete comments too?

tests/api/user/delete.test.coffee
maxlath: it's not possible at the moment, neither before or after this PR, but having a look at it and the implications could be a todo

server/controllers/items/lib/delete_user_items.coffee
maxlath: not sure of what you mean

jum-s: garbage collect orphan comments

maxlath: there are no more items comments in the app, trying to delete those was thus legacy code that this PR removes

jum-s: ok

PR #235, opened 2018-12-03
cleaning ttl dump

maxlath:
* validating dumps ecd6b9e 
* update date formatter to take into account month precision da9dd98
* added support for invp:P2 claims 035f1b1, 4ad003e


scripts/dumps/lib/serialize_entity_in_turtle.coffee
```diff
@@ -2,13 +2,15 @@ CONFIG = require 'config'
...
+{ yellow } = require 'chalk'
 
 module.exports = (entity)->
   { _id } = entity
 
   text = "inv:#{_id} a wikibase:Item ;"
 
   for lang, value of entity.labels
+    value = formatStringValue value
     text += """\n  rdfs:label #{formatStringValue(value)}@#{lang} ;"""
```
jum-s: `formattedValue` instead of `value` ? which can then be used as : 
```
text += """\n  rdfs:label #{formattedValue}@#{lang} ;"""
```
and instead of formattedValue which is very generic why not using `slug`

maxlath: isn't `slug` specific to URLs?

jum-s: `label` then ? something more specific than value?

maxlath: addressed in c7cc70cb

Issue #67, opened 2016-04-15
Move translation out of Transifex


frimelle:
Might be considered to use https://translatewiki.net/ for the translation of inventaire to have it there with a lot of other cool projects for people to join the translation efforts.

maxlath:
@frimelle interesting! I love the fact that anyone can contribute without creating an account first; the typical workflow on Transifex being:
- someone wants to help translating and find the Transifex project, create an account and make a request
- I see the notification one week later and validate the request to join a given language team
- the person long forgot about it and, in the case she got a notification for the acceptance, is probably not in the flow any more

A few questions:
- do you know how the API workflow works? that's probably the biggest obstacle: re-integrating strings imports and exports
- are the strings mutualized between the projects? like, if a string is not translated in inventaire in Dutch but the same string is available in Dutch in an other project, it will be imported? that would be beautiful :D

almereyda:
We may offer [weblate.allmende.io](https://weblate.allmende.io) at some point https://github.com/TransforMap/weblate.transformap.co/issues/1

maxlath:
we are well installed on Transifex at the moment, and despite it's non-opensourceness, the cost of moving to another platform is too high: closing the issue for now

maxlath:
https://github.com/WeblateOrg/weblate

almereyda:
If you please we could migrate weblate.transformap.co to weblate.allmende.io
and offer that one instead.

almereyda:
@maxlath Newest Weblate 2.20 is now running on https://weblate.allmende.io

Just sayin'.

maxlath:
re-opening as Transifex might be hurting contributions by making progresses hard to follow (maybe contributors stats are a premium?!?)
poke @jum-s

maxlath:
requirements for the new translation platform:
* libre
* clear interface
* translations can be extracted from an API, preferably without authentification
* statistics on contributions and evolution of translations, so that we can see when there is a big effort going on, offer help, and thank the contributor

anything else?

maxlath:
some [input from @wetneb](https://mamot.fr/@pintoch/102121968201990744):
> In my experience Weblate is just fantastic, it feels very professional. And Translatewiki brings you a lot of very keen translators who will translate your app in languages you had no idea existed!

nemobis:
This is how to add a project on translatewiki.net:
https://translatewiki.net/wiki/Translating:New_project

We can work on it now that we're in Stockholm, happy to help.

[...]

PR #236, opened 2018-12-08
refactor i18n by gathering i18n strings and scripts in a dedicated repository

maxlath:
**Previous situation**
* i18n scripts splitted between server and client repos
* i18n strings hidden on their own orphan-ish branch: [`i18n`](https://github.com/inventaire/inventaire/tree/i18n)

**After this refactor**
* all i18n scripts and strings gathered in a common repo: [inventaire/inventaire-i18n](https://github.com/inventaire/inventaire-i18n)

The workflow to make it work smoothly might need some more ajustements, but I think that's a sane base to work from.

**See also**
client side PR https://github.com/inventaire/inventaire-client/pull/161


maxlath:
this, btw, should make moving to another translation interface easier (see #67)

PR #243, opened 2019-01-01
removing alias uris and sitelinks resolution support

maxlath:
expert users wanting to use this kind of feature can now use https://tools.wmflabs.org/hub/

Ex: https://tools.wmflabs.org/hub/frwiki:Lucien_Suel?site=inventaire => https://inventaire.io/entity/wd:Q3265721


maxlath:
remerged after rebase on master in f030d479

Issue #182, opened 2018-05-22
remove Wikidata edition items from quarantine


maxlath:
Wikidata editions are currently ignored for 2 reasons:
* editions and works distinction in Wikidata is currently a mess, with works holding edition-level properties (especially identifiers such as ISBNs)
* edition in Wikidata can't link to cover images, which are critical to Inventaire: this could be worked around with local shadow items (#181)

maxlath:
Done in https://github.com/inventaire/inventaire/pull/753 :tada: 

It still excludes editions that we consider properly shaped: editions not associated to a work, or missing a title (`wdt:P1476`)

PR #251, opened 2019-01-06
reverse_claims: run requests on local entities only for certain properties

maxlath:
typically for editions properties, given that editions from Wikidata are still ignored
see #182


jum-s:
this PR makes a server responsability to know if a property should be queried on wikidata or only locally.

PR #253, opened 2019-01-07
entities delete by uris: fix issue with items not detected

maxlath:
due to a different URI being used by item.entity, typically an ISBN, while the `inv:` URI was passed


maxlath:
remerged in dd6a6010

PR #256, opened 2019-01-09
perf refactor : query byClaimsValue only once

jum-s:
some may say '.tap' is not that elegant or might cause problems, but
hey test are green...


maxlath:
so the behavior would now be that it deletes until an entity can't be deleted, in which case it stops there, with part of the entities removed, one denied, and the rest ignored. I think we should either have all the current behavior of all-or-nothing, or, all entities that can be removed are removed, the failing ones being ignored (but listed in some way in the response)

PR #258, opened 2019-01-10
Fix recent public items

jum-s:
uses `publicByDate` couch view in order to fill recent items as `publicByLangAndDate` is now obsolete

### DIFF
--- a/server/controllers/items/recent_public.coffee
+++ b/server/controllers/items/recent_public.coffee
@@ -5,7 +5,6 @@ items_ = __.require 'controllers', 'items/lib/items'
 sanitize = __.require 'lib', 'sanitize/sanitize'
 bundleOwnersToItems = require './lib/bundle_owners_to_items'
 itemsQueryLimit = 100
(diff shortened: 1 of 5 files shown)


server/controllers/items/recent_public.coffee
```diff
@@ -22,20 +22,28 @@ module.exports = (req, res)->
...
-  # wrapping lodash .chain() function
-  .value()
+selectRecentItems = (lang, limit)-> (items)->
+  recentItems = []
+  discardedItems = []
+  itemsCountByOwner = {}
+
+  for item in items
+    if recentItems.length is limit then return recentItems
+    if item.snapshot['entity:lang'] is lang and itemsCountByOwner[item.owner] < 3
```
maxlath: if `itemsCountByOwner[item.owner]` is `undefined`, `itemsCountByOwner[item.owner] < 3` is `false`, so initialization (`itemsCountByOwner[item.owner] ?= 0`) should be done before this line

PR #260, opened 2019-01-17
sanitize endpoint exists or create from from seed

jum-s:
add authors and isbn as sanitize parameters
add tests with isbns sample in fixtures


tests/api/fixtures/entities.coffee
```diff
@@ -29,6 +34,7 @@ module.exports = API =
   createHuman: createEntity 'wd:Q5'
   createWork: createEntity 'wd:Q571'
   createSerie: createEntity 'wd:Q277759'
+  createIsbn: -> _.sample someIsbns
```
maxlath: that's not really a creation, what about renaming it `getSomeIsbn`?

PR #261, opened 2019-01-17
Archive removed placeholder tasks

maxlath:
to avoid have old tasks hanging around while their suspect entity was removed already


maxlath:
remerged after rebase on master in a4df30c6

PR #262, opened 2019-01-18
items:update-bulk update one attribute on many items

jum-s:
open endpoint `bulk-update` for smoothing transition out of `update` and not breaking the api
import controller and model logic from [tasks](server/server/models/task.coffee)

### DIFF
--- a/server/models/validations/item.coffee
+++ b/server/models/validations/item.coffee
@@ -11,6 +11,7 @@ module.exports = validations =
   entity: entityUri
   lang: (lang)-> if lang then _.isLang(lang) else true
   pictures: (pictures)-> _.isArray(pictures) and _.every(pictures, imgUrl)
+  attribute: (attribute)-> attribute in _.keys(constrained)
   transaction: (transaction)->
     transaction in constrained.transaction.possibilities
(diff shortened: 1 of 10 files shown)


jum-s:
when client is ready, replace `update` by  `bulk-item`

server/models/validations/item.coffee
```diff
@@ -11,6 +11,7 @@ module.exports = validations =
   entity: entityUri
   lang: (lang)-> if lang then _.isLang(lang) else true
   pictures: (pictures)-> _.isArray(pictures) and _.every(pictures, imgUrl)
+  attribute: (attribute)-> attribute in _.keys constrained
```
maxlath:
```suggestion
  attribute: (attribute)-> attribute in _.keys(constrained)
```

maxlath:
I added f3d5b66 to make the Item model test pass. For me, this can either be merged now, or wait for the client [`inventory-browser` branch](https://github.com/inventaire/inventaire-client/commits/inventory-browser) to be merged to fully replace `/api/items?action=update` with that new `bulk-update` endpoint

PR #264, opened 2019-01-24
add entities timestamps

maxlath:
to easily be able to access creation and last update dates for data cleaners, or from query.inventaire.io

### DIFF
--- a/server/models/entity.coffee
+++ b/server/models/entity.coffee
@@ -38,6 +38,7 @@ module.exports = Entity =
     type: 'entity'
     labels: {}
     claims: {}
+    created: Date.now()
 
   setLabel: (doc, lang, value)->
     assert_.types [ 'object', 'string', 'string' ], arguments
@@ -53,6 +54,9 @@ module.exports = Entity =
       throw error_.new 'already up-to-date', 400, { doc, lang, value }
 
     doc.labels[lang] = value
+
+    doc.updated = Date.now()
+
     return doc
 
   setLabels: (doc, labels)->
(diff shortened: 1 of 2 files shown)


maxlath:
the map reduce that could be run on patches to find the `created` and `updated` timestamps of pre-existing entities:
```json
{
   "_id": "_design/patches_migration",
   "language": "coffeescript",
   "views": {
       "entities_timestamps": {
           "map": "(doc)->\n  entityId = doc._id.split(':')[0]\n  emit entityId, doc.timestamp",
           "reduce": "(keys, values, rereduce)->\n  min = Infinity\n  max = 0\n  values.forEach (value)->\n    if value < min then min = value\n    if value > max then max = value\n  return { created: min, updated: max }\n  "
       }
   }
}
```

server/models/entity.coffee
```diff
@@ -38,6 +38,7 @@ module.exports = Entity =
     type: 'entity'
     labels: {}
     claims: {}
+    created: Date.now()
```
jum-s: for consistency i would add `updated: Date.now()` field too when entity is created, so that every entity has a `updated` field no matter what

tests/unit/models/024-entity.coffee
```diff
@@ -208,3 +264,152 @@ describe 'entity model', ->
...
+        entityB.claims['wdt:P921'].should.deepEqual [ 'wd:Q1', 'wd:Q3' ]
+        done()
+
+      it 'should not create duplicated claims', (done)->
+        entityA = workDoc()
+        entityB = workDoc()
+        Entity.createClaim entityA, 'wdt:P921', 'wd:Q3'
+        Entity.createClaim entityB, 'wdt:P921', 'wd:Q3'
+        Entity.mergeDocs entityA, entityB
+        entityB.claims['wdt:P921'].should.deepEqual [ 'wd:Q3' ]
```
jum-s: +1

server/models/entity.coffee
maxlath: done in 3caa775

PR #266, opened 2019-01-25
entities merge: reject circular merges

maxlath:


server/controllers/entities/merge.coffee
```diff
@@ -58,6 +58,9 @@ merge = (reqUserId, toPrefix, fromUri, toUri)-> (res)->
   toEntity = entities[toUri] or entities[redirects[toUri]]
   unless toEntity? then throw notFound 'to', toUri
 
+  if fromEntity.uri is toEntity.uri
+    throw error_.new 'circular merge', 400, { fromUri, toUri }
```
jum-s: as circular can be ambiguous term (looping) what about :
`'merging to the same entity' , 400, { toUri }`

maxlath: replaced by `can't merge an entity into itself`

PR #267, opened 2019-01-25
entities: improve update-claims error message and cleanup tests

maxlath:


tests/api/entities/update_claims.test.coffee
```diff
@@ -50,14 +50,28 @@ describe 'entities:update-claims', ->
 
     return
 
+  it 'should reject an update on an unexisting claim', (done)->
+    hash1 = '2e31caf289888dba0e88944c98628d4296d431ee'
+    hash2 = '3d10239b31d46e9c1fe288de376621b8bcc00051'
```
jum-s: kis? what about casual stable day-to-day claim like a `P31:Q5` or a `P212:Qx` fanciness is the enemy of simplicity

jum-s: yay!

PR #268, opened 2019-01-25
server startup: gain a few ms and cleanup

maxlath:


server/init_express.coffee
```diff
@@ -4,30 +4,34 @@ _ = __.require('builders', 'utils')
...
+console.time 'middlewares'
 middlewares = require './middlewares/middlewares'
 middlewaresList = middlewares.common.concat (middlewares[CONFIG.env] or [])
 
 middlewareErrorHandler = require './middlewares/error_handler'
+console.timeEnd 'middlewares'
 
+console.time 'routes'
 routes = require './controllers/routes'
+console.timeEnd 'routes'
```
jum-s: is there a need to leave those benchmark tool displaying at every start up ?

jum-s:
+1 for documentation

server/init_express.coffee
maxlath: removed

PR #272, opened 2019-02-10
Entry and entries resolver

jum-s:
first attemps of resolving external sources entries
with two endpoints `resolve` and `resolve-entries`

throwing on isbn missing in edition
can take options as parameters:
`create` for creating new entities from entry
`update` for updating existing entity claims (without adding new values if prop already exists)


jum-s:
:100:

Issue #155, opened 2018-03-09
allow to create publisher entities


maxlath:
all publishers aren't in Wikidata, far from it regarding small publishers or big publishers sub-branches

publishers properties:
```sh
'wdt:P31': [ "wd:Q2085381" ]
'wdt:P571' # inception date
'wdt:P127' # owned by
'wdt:P159' # headquarters location
'wdt:P3035' # ISBN publisher prefix
'wdt:P856' # official website
'wdt:P2002' # Twitter account
'wdt:P2013' # Facebook account
'wdt:P2003' # Instagram username
'wdt:P2397' # YouTube channel ID
'wdt:P4033' # Mastodon address
```

olivierberten:
+1

maxlath:
[server](https://github.com/inventaire/inventaire/pull/275) and [client](https://github.com/inventaire/inventaire-client/pull/169) PRs merged and online!

PR #275, opened 2019-02-25
[issue #155] Create inventaire publishers

jum-s:
minimal attempt of creating and updating inventaire publishers
addressing #155

### DIFF
--- a/server/controllers/entities/lib/properties/properties_values_constraints.coffee
+++ b/server/controllers/entities/lib/properties/properties_values_constraints.coffee
@@ -48,6 +48,8 @@ module.exports =
   'wdt:P136': bases.entity
   # based on
   'wdt:P144': bases.workEntity
+  # headquarters location
+  'wdt:P159': bases.entity
   # serie
   'wdt:P179': bases.serieEntity
   # ISNI
@@ -76,8 +78,6 @@ module.exports =
   'wdt:P570': bases.uniqueSimpleDay
   # inception
   'wdt:P571': bases.uniqueSimpleDay
-  # dissolution date
-  'wdt:P576': bases.uniqueSimpleDay
   # publication date
   'wdt:P577': bases.uniqueSimpleDay
   # edition of
@@ -136,7 +136,7 @@ module.exports =
   'wdt:P2963': builders.externalId /^[1-9]\d*$/
   # GoodReads book ID
   'wdt:P2969': builders.externalId /^[1-9]\d*$/
-  # ISBN publisher
+  # ISBN publisher prefix
   'wdt:P3035': builders.externalId /^978-\d-\d{1,}$/
   # Babelio author ID
   'wdt:P3630': builders.externalId /^\d+$/
--- a/server/controllers/entities/lib/properties/properties_values_constraints.coffee
+++ b/server/controllers/entities/lib/properties/properties_values_constraints.coffee
@@ -30,12 +30,18 @@ extend = (base, extension)-> _.extend {}, base, extension
 module.exports =
   # image
   'invp:P2': bases.imageHash
+  # country
+  'wdt:P17': bases.entity
   # instance of
   'wdt:P31': extend bases.uniqueEntity, { adminUpdateOnly: true }
   # author
   'wdt:P50': extend bases.humanEntity, { hasPlaceholders: true }
+  # founded by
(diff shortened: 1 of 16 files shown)


server/controllers/entities/lib/properties/properties_values_constraints.coffee
```diff
@@ -65,6 +73,8 @@ module.exports =
   'wdt:P569': bases.uniqueSimpleDay
   # date of death
   'wdt:P570': bases.uniqueSimpleDay
+  # inception
+  'wdt:P571': bases.uniqueSimpleDay
```
maxlath: it looks like fae5f7c mistakenly removed `wdt:P576` or am I missing something?

server/controllers/entities/lib/properties/properties_config_bases.coffee
```diff
@@ -25,6 +25,7 @@ module.exports =
   workEntity: restrictedEntityType 'work'
   serieEntity: restrictedEntityType 'serie'
   humanEntity: restrictedEntityType 'human'
+  countryEntity: restrictedEntityType 'countrie'
```
maxlath: what about using `country` as the singular form? I think the added cost is acceptable, see proposed implementation d5abf5a

server/lib/wikidata/aliases.coffee
```diff
@@ -82,14 +82,20 @@ module.exports = typesAliases =
     # bandes dessinées publishing company
     'wd:Q3279251'
   ]
-
+  # Types required to update the entity search engine
+  # see server/controllers/entities/lib/update_search_engine
```
maxlath: this comment should indeed be here now, but should removed above `publishers`

server/controllers/entities/lib/properties/properties_values_constraints.coffee
jum-s: not really mistakenly, i decided it wasnt a relevant information enough to be displayed so i deleted without explaining my motivation...

server/controllers/entities/lib/properties/properties_config_bases.coffee
jum-s: it hurts my eyes less to see `countrie` here, than a `if pluralType is 'countries' then 'country'` 
its a hack but at least it stays with the other hardcoded types above

server/controllers/entities/lib/properties/properties_values_constraints.coffee
jum-s: [for documentation purpose] could actually be useful for check publication editions from a post-mortem editor

server/controllers/entities/lib/properties/properties_values_constraints.coffee
```diff
@@ -31,14 +31,20 @@ module.exports =
...
+  # owned by
+  'wdt:P127': bases.entity
   # movement
   'wdt:P135': bases.entity
   # genre
   'wdt:P136': bases.entity
   # based on
   'wdt:P144': bases.workEntity
+  # headquarters location
+  'wdt:P159': bases.entity
```
maxlath: this should probably also be let aside (as for `wdt:P17`) as we can't properly validate a location entity at the moment

server/controllers/entities/lib/properties/properties_values_constraints.coffee
```diff
@@ -123,6 +133,8 @@ module.exports =
   'wdt:P2963': builders.externalId /^[1-9]\d*$/
   # GoodReads book ID
   'wdt:P2969': builders.externalId /^[1-9]\d*$/
+  # ISBN publisher prefix
+  'wdt:P3035': builders.externalId /^978-\d-\d{1,}$/
```
maxlath: allow `979` prefixes and be more restrictive on the following blocks (copying [P3035#P1793](https://www.wikidata.org/wiki/Property:P3035#P1793) regex)
```suggestion
  'wdt:P3035': builders.externalId /^97(8|9)-\d{1,5}-\d{2,7}$/
```

server/controllers/entities/search_type.coffee
```diff
@@ -20,6 +20,7 @@ indexedTypes = [
   'movements'
   'publishers'
   'collections'
+  'countries'
```
maxlath: `countries` can be removed for now

PR #282, opened 2019-03-13
Dumps

maxlath:
## entities dump
add `schema:dateCreated` (from `entity.created`)
add `schema:dateModified` (from `entity.updated`)
add `schema:version` (from `entity._rev`)
add `wikibase:statements` (statements count)
add `wikibase:labels` (labels count)
remove `skos:prefLabel` (duplicated with `rdfs:label`)

## items dump
sketched but commented-out as it doesn't pass the validation and might not be a good idea after all (see [privacy concerns](https://github.com/inventaire/inventaire/blob/9a9b874/scripts/dumps/prepare_dumps#L4-L12))


scripts/dumps/lib/serialize_entity_in_turtle.coffee
```diff
@@ -5,16 +5,31 @@ properties = __.require 'controllers', 'entities/lib/properties/properties_value
...
 
+  dateCreated = new Date(created).toISOString()
+  text += """\n  schema:dateCreated "#{dateCreated}"^^xsd:dateTime ;"""
+
+  dateModified = new Date(updated).toISOString()
+  text += """\n  schema:dateModified "#{dateModified}"^^xsd:dateTime ;"""
+
+  # Using the _rev to deduce the version might give a version number
+  # higher than the real number of patches, but that's better than nothing
+  version = parseInt _rev.split('-')
```
jum-s: `parseInt` indeed according to the manual "If parseInt encounters a character that is not a numeral in the specified radix, it ignores it and all succeeding characters and returns the integer value parsed up to that point." its not it's primary function tho. why not go with a `_rev.split('-')[0]` ?

scripts/dumps/lib/serialize_entity_in_turtle.coffee
```diff
@@ -23,6 +38,13 @@ module.exports = (entity)->
     else
       console.warn yellow('missing formatter'), datatype
 
+  text += """\n  wikibase:statements #{statementsCount} ;"""
```
jum-s: maybe `"""\n  wikibase:statements count #{statementsCount} ;"""` ? to be clear about what is this integer exactly

scripts/dumps/lib/serialize_entity_in_turtle.coffee
```diff
@@ -23,6 +38,13 @@ module.exports = (entity)->
...
+
+  labelsCount = Object.keys(entity.labels).length
+  # This property isn't actually used by Wikidata
+  # but builds on the idea of 'wikibase:statements'
+  text += """\n  wikibase:labels #{labelsCount} ;"""
```
jum-s: idem

scripts/dumps/prepare_items_dumps
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
+source ./scripts/dumps/lib/utils.sh
+
+items_dump="${today_folder}/items.ttl"
+dbHost=$(node -p "require('config').db.fullHost()")
+
+echo "
+@prefix wdt: <http://www.wikidata.org/prop/direct/> .
+@prefix inv: <http://inventaire.io/entity/> .
+@prefix invexamplar: <http://inventaire.io/items/> .
```
jum-s: wuhu what a brand new name ha ?!

jum-s: maybe `invex` as its should stay a short string as a prefix ?

scripts/dumps/prepare_items_dumps
```diff
@@ -0,0 +1,21 @@
...
+" > $items_dump
+
+# We only need the items entities for now
+curl -s "${dbHost}/items-prod/_design/items/_view/byEntity" |
+  jq '.rows[] | [.id, .key[0]] | join(" wdt:P1574 ")' |
```
jum-s: arent you pushing item ID in the ttl with this line ?

scripts/dumps/prepare_items_dumps
maxlath: yeah that was temporary, and with the whole items dump being commented-out now, that could even be removed

scripts/dumps/prepare_items_dumps
maxlath: yes, that's why it's [commented-out in `prepare_dumps`](https://github.com/inventaire/inventaire/blob/9a9b8744795c738949fee99d05689576ae4b3a0b/scripts/dumps/prepare_dumps#L4-L14), but maybe that should just be removed(?) Keeping track of it somewhere could be of value though

scripts/dumps/lib/serialize_entity_in_turtle.coffee
maxlath: I don't get it, what is the role of that `count` keyword in the triple?

scripts/dumps/prepare_items_dumps
maxlath: renamed in c556bb8e

Issue #90, opened 2017-05-18
Different books link to the same book


thewholelifetolearn:
I have encountered a big issue. I was going through my inventory to update the information of my books when I saw on https://inventaire.io/entity/inv:aea555cb4000c73f30e5288af10e16c8 that there were several book seen as editions link to one book.

In this case, the volumes 12, 13, 14 et 15 of Bakuman are seen as editions of the volume 12. What is expected is to have a book for each volume.

maxlath:
this is weird, those editions should indeed have one associated work each. The reasons of this issue I can think of:
* the server mistakenly assumed that those similar title where refering to several editions of a unique work
* a user mistakenly added editions to this work, assuming this was the Bakuman series page

It shouldn't be too hard to fix it, but can't be done from the user interface, I will do this in the coming days.

This kind of issues are to be expected given the way our knowledge base is build - with a mix of server assumptions and open human contributions -, we should just progressively get better at tackling those, making them less likely, and easier to fix :)

thewholelifetolearn:
If it is a user mistake, it could be because there is no big difference display between series and a work. Even with an edition.

Perhaps, should there be a manager kind of users. They could correct things, manage duplicates, etc.. That would give some hassle less for the administrator (in this case, you).

thewholelifetolearn:
There is the same issue with the volumes 3 and 9: https://inventaire.io/entity/inv:aea555cb4000c73f30e5288af10cbe01

thewholelifetolearn:
Same thing with volumes 36 and 37 of "Prince du Tennis": https://inventaire.io/entity/inv:aea555cb4000c73f30e5288af1091208

maxlath:
I fixed those different cases but we should probably let this issue opened in case more come and until we can get it fixed: I'm now pretty sure that it's the server that is too aggressive at matching editions to a common work when the title and the authors names match; it works for most books but visibly not for manga and series where works are more often titled on the pattern "serie name - volume x"

thewholelifetolearn:
Indeed, I don't know what are the sources of the information, but in the cases of comic books series or manga series, the names are often "series name -  series name" or just "series name".  
That is the reason I'm renaming a lot of them to "series name - volume[ - name]".

Thank you for making the changes. I will let the issue open if the case happens again.

thewholelifetolearn:
[...]

PR #285, opened 2019-03-23
work_entity_search_utils: matchTitle: be more strict on title matching for volume titles

maxlath:
so that all volume editions don't end up being linked to the same work

addressing #90


jum-s:
I thought that `scaffold_entity_from_seed` would be replaced by the `resolver` soon.
So I will merge quickly and pretend i never saw this PR :)

maxlath:
rebased and remerged in f94aeac3 to keep the history clean

PR #286, opened 2019-03-23
 search: allow to customize the limit

maxlath:
before those changes, whatever the request limit, only 20 results where taken from ElasticSearch

### DIFF
--- a/tests/api/search/search.test.coffee
+++ b/tests/api/search/search.test.coffee
@@ -37,18 +37,7 @@ describe 'search:global', ->
     .then undesiredRes(done)
     .catch (err)->
       err.statusCode.should.equal 400
-      err.body.status_verbose.should.equal 'invalid type: da'
-      done()
-    .catch undesiredErr(done)
-
-    return
-
-  it 'should reject search without lang', (done)->
-    nonAuthReq 'get', '/api/search?search=yo&types=works'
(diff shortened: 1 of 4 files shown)


tests/api/search/search.test.coffee
```diff
@@ -37,18 +37,7 @@ describe 'search:global', ->
     .then undesiredRes(done)
     .catch (err)->
       err.statusCode.should.equal 400
-      err.body.status_verbose.should.equal 'invalid type: da'
-      done()
-    .catch undesiredErr(done)
-
-    return
-
-  it 'should reject search without lang', (done)->
```
jum-s: how is this test not relevant anymore?

maxlath: the use of sanitize [makes the `lang` parameter default to `en`](https://github.com/inventaire/inventaire/blob/daa39bc/server/lib/sanitize/parameters.coffee#L111), so 3 options:
* remove the spec to throw on a missing lang parameter (the one currently implemented)
* customize the parameter config to request that a missing `lang` be rejected
* change the parameter general behavior, but that might impact other places depending on it

jum-s: what about testing the default lang behavior ?
`should return a english lang label if no language parameter is passed`

PR #294, opened 2019-04-29
[WIP] get_entity_type: enforce disagreements with Wikidata P31 typing

maxlath:
especially,
- an entity can't be an edition if it doesn't have `wdt:P629` claims
- prefer to consider an ambiguous entity as a serie rather than a work
  as it will be then easier to create subparts of this entity to which to link editions
  In the worst case, that was actually not a serie and subparts can be merged into it


server/controllers/entities/lib/get_entity_type.coffee
```diff
@@ -2,23 +2,49 @@ __ = require('config').universalPath
...
+        unless claims['wdt:P629']? then return 'work'
+      return type
+    else
+      # Case with multiple P31 values: try to guess the most credible type
+      isWork = 'work' in propertyTypes
+      isEdition = 'edition' in propertyTypes
+      isSerie = 'serie' in propertyTypes
+
+      # If it has 'edition of' claims, consider that it's an edition
+      if isEdition and claims['wdt:P629']? then return 'edition'
```
jum-s: We might consider returning a `undefined` value as no wikidata edition are to be interpreted by inventaire has an edition. Inventaire would then consider this entity as a topic, not an edition, avoiding for example the possibility of adding it to a user's inventaire.

jum-s (changes requested):

jum-s:
Some code to [refactor](https://github.com/inventaire/inventaire/blob/cd065234d702560ac865e0a2a6c5363c25314517/server/controllers/entities/lib/format_edition_entity.coffee#L8)

maxlath:
@jum-s how would you like to see [it](https://github.com/inventaire/inventaire/blob/cd065234d702560ac865e0a2a6c5363c25314517/server/controllers/entities/lib/format_edition_entity.coffee#L8) refactored?

maxlath:
@jum-s as you know I'm having doubts about the whole thing, please do not merge yet

PR #296, opened 2019-05-04
Add task occurences authorities

jum-s:
- improve tasks occurences by querying some national libraries author works title
- automerge only if some of those libraries domain name are in occurences urls

### DIFF
--- a/server/data/lib/fetch_external_author_works_titles.coffee
+++ b/server/data/lib/fetch_external_author_works_titles.coffee
@@ -1,0 +1,18 @@
+CONFIG = require 'config'
+__ = CONFIG.universalPath
+_ = __.require 'builders', 'utils'
+requests_ = __.require 'lib', 'requests'
+qs = require 'querystring'
+
+module.exports = (endpoint, query)->
+  escapedQuery = qs.escape query
+  base = "#{endpoint}?query="
+  headers = { accept: 'application/sparql-results+json' }
+  url = base + escapedQuery
+
+  requests_.get { url, headers }
+  .then (res)->
+    res.results.bindings
+    .map (result)->
+      quotation: result.title?.value
+      url: result.work?.value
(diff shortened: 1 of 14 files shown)


server/data/lib/fetch_external_author_works_titles.coffee
```diff
@@ -0,0 +1,18 @@
+CONFIG = require 'config'
+__ = CONFIG.universalPath
+_ = __.require 'builders', 'utils'
+requests_ = __.require 'lib', 'requests'
+qs = require 'querystring'
+
+module.exports = (endpoint, query)->
```
maxlath: if every reuser of that function needs to call `cache_.get`, that could be DRYed a bit further:
```coffee
# fetch_external_author_works_titles.coffee
CONFIG = require 'config'
__ = CONFIG.universalPath
_ = __.require 'builders', 'utils'
requests_ = __.require 'lib', 'requests'
qs = require 'querystring'
cache_ = __.require 'lib', 'cache'
{ oneMonth } =  __.require 'lib', 'times'
timespan = 3 * oneMonth

module.exports = (name, endpoint, getQuery)-> (id)->
  key = "#{name}:author-works-titles:#{id}"
  return cache_.get { key, fn: fetch.bind(null, endpoint, getQuery, id), timespan  }

fetch = (endpoint, getQuery, id)->
  escapedQuery = qs.escape getQuery(id)
  base = "#{endpoint}?query="
  headers = { accept: 'application/sparql-results+json' }
  url = base + escapedQuery

  requests_.get { url, headers }
  .then (res)->
    res.results.bindings
    .map (result)->
      quotation: result.title?.value
      url: result.work?.value
 ```
Then the consumers would just have to specify what's missing:
```coffee
# get_bne_author_works_titles.coffee
CONFIG = require 'config'
__ = CONFIG.universalPath
fetchExternalAuthorWorksTitles = __.require 'data', 'lib/fetch_external_author_works_titles'

endpoint = 'http://datos.bne.es/sparql'

getQuery = (bneId)->
  """
  SELECT ?work ?title WHERE {
    <http://datos.bne.es/resource/#{bneId}> <http://datos.bne.es/def/OP5001> ?work .
    ?work <http://datos.bne.es/def/P1001> ?title .
  }
  """

module.exports = fetchExternalAuthorWorksTitles 'bne', endpoint, getQuery
```

PR #303, opened 2019-06-07
search filter: accept an entity prefix as filter parameter

maxlath:
this would make [query.inventaire.io](https://query.inventaire.io) prefix autocompletion more efficient: it currently needs to request a lot of results to then filter-out those that aren't of the `inv` prefix

### DIFF
--- a/server/controllers/search/lib/type_search.js
+++ b/server/controllers/search/lib/type_search.js
@@ -2,20 +2,33 @@ const CONFIG = require('config')
 const __ = CONFIG.universalPath
 const requests_ = __.require('lib', 'requests')
 const assert_ = __.require('utils', 'assert_types')
+const error_ = __.require('lib', 'error/error')
+const { Promise } = __.require('lib', 'promises')
 const { host: elasticHost } = CONFIG.elasticsearch
 const { formatError } = __.require('lib', 'elasticsearch')
 const getIndexesAndTypes = require('./get_indexes_and_types')
 const queryBodyBuilder = require('./query_body_builder')
 
 // types should be a subset of ./types possibleTypes
-module.exports = (types, search, limit = 20) => {
-  let indexes
+module.exports = (types, search, limit = 20, filter) => {
   assert_.array(types)
-  assert_.string(search);
+  assert_.string(search)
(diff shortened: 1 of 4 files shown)


tests/api/search/search.test.coffee
```diff
@@ -45,6 +45,28 @@ describe 'search:global', ->
...
+        .then undesiredRes(done)
+        .catch (err)->
+          err.statusCode.should.equal 400
+          err.body.status_verbose.should.startWith 'invalid filter: foo'
+          done()
+        .catch undesiredErr(done)
+
+        return
+
+      it 'should accept a filter parameter', (done)->
```
jum-s: Would be more complete to test with a fixtured entity with a wd label 
`author = createHuman { labels: { en: 'Gilles Deleuze' } }`
and check its presence 
`search 'humans', 'Gilles Deleuze', 'inv'`
`_.map(results, 'uri').includes(author.uri).should.be.true()`
and its absence 
`search 'humans', 'Gilles Deleuze', 'wd'`
`_.map(results, 'uri').includes(author.uri).should.be.false()`
well doesnt have to be Gillou, could be a random label

maxlath: yep, that was definitely a good idea: the test was passing but no search was done ^^ patch incoming

maxlath:
so after rebasing and fixing the test as recommanded, we have a problem: the test passes in some cases (typically when the database isn't empty), but not in others (typically when running with an empty database), as exactly matching labels fail to earn more that the [currently set `min_score=0.5`](https://github.com/inventaire/inventaire/blob/cfe35fa/server/controllers/search/lib/query_body_builder.js#L12v).  (and when not empty, the hits scores aren't going much higher than 0.5). I'm not sure the reason why ElasticSearch would return such a low score for exact matching label, but I'm afraid we will have to dig into that again at some point, this might also be the reason why other search tests periodically fail

tests/api/search/search.test.coffee
maxlath: done

maxlath:
made obsolete by #462

PR #331, opened 2019-11-07
queries: harmonize work type detection

maxlath:
hypothesis of this PR: using `VALUES` allows to more easily see the pattern than making those lenghtly `UNION`s

### DIFF
--- a/server/data/wikidata/queries/author_works.js
+++ b/server/data/wikidata/queries/author_works.js
@@ -2,6 +2,14 @@ module.exports = {
   parameters: [ 'qid' ],
   query: params => {
     const { qid: authorQid } = params
+
+    // P50    author
+    // P58    screenwriter
+    // P110   illustrator
+    // P6338  colorist
+
+    // Filter-out instances of edition (Q3331189)
+
     return `SELECT ?work ?type ?date ?serie WHERE {
   ?work wdt:P50|wdt:P58|wdt:P110|wdt:P6338 wd:${authorQid} .
   ?work wdt:P31 ?type .
--- a/server/data/wikidata/queries/editions_reverse_claims.js
+++ b/server/data/wikidata/queries/editions_reverse_claims.js
@@ -2,10 +2,20 @@ module.exports = {
   parameters: [ 'pid', 'qid' ],
   query: params => {
     const { pid, qid } = params
+    // Filter-out entities tagged as both work and edition (Q3331189)
+
+    // Work types:
+    // Q571       book
+    // Q47461344  written work
+    // Q7725634   literary work
+    // Q2831984   comic book album
+    // Q1004      comic
+    // Q1760610   comic book
+    // Q8274      manga
+
     return `SELECT DISTINCT ?item WHERE {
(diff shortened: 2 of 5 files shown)


server/data/wikidata/queries/editions_reverse_claims.coffee
```diff
@@ -7,7 +7,10 @@ module.exports =
       ?item wdt:#{pid} wd:#{qid} .
       ?item wdt:P31 wd:Q3331189 .
       # Filter-out entities tagged as both work and edition
-      FILTER NOT EXISTS { ?item wdt:P31 wd:Q571 }
+      FILTER NOT EXISTS {
+        ?item wdt:P31 ?work_type
+        VALUES (?work_type) { (wd:Q571) (wd:Q47461344) (wd:Q7725634) (wd:Q2831984) (wd:Q1004) (wd:Q1760610) (wd:Q8274) } .
```
jum-s: A usual pattern is to define a variable before using it, but here `?work_type` is defined after.

server/data/wikidata/queries/works_reverse_claims.coffee
```diff
@@ -5,26 +5,8 @@ module.exports =
...
-      # comic book series
-      UNION { ?item wdt:P31 wd:Q14406742 . }
-      # manga series
-      UNION { ?item wdt:P31 wd:Q21198342 . }
-      # novel series
-      UNION { ?item wdt:P31 wd:Q1667921 . }
-      # Filter-out entities tagged as both work and edition
-      FILTER NOT EXISTS { ?item wdt:P31 wd:Q3331189 }
+      ?item wdt:P31 ?work_type .
+      VALUES (?work_type) { (wd:Q571) (wd:Q47461344) (wd:Q7725634) (wd:Q2831984) (wd:Q1004) (wd:Q1760610) (wd:Q8274) } .
```
jum-s: before ther was some comment on what Qids meant, which made it more readable

jum-s (changes requested):
unsure if its more readable, this `VALUES`  in SPARQL is new to us but it does not make it more readable for everyone. I would claim `UNION` is more explicit as its used in many SQL implementations

server/data/wikidata/queries/works_reverse_claims.coffee
maxlath: done in befb87f8

server/data/wikidata/queries/editions_reverse_claims.coffee
maxlath: done in 04397e42

maxlath:
I rewrote the commit for the post-decaffeinate era

maxlath:
on the `VALUES` version `UNION` syntax debate: I now generally prefer the `VALUES` syntax, as, when you are familiar with it, it is much more explicit on what is happening than when you have to inspect all `UNION` clauses to then figure out that it is the exact same pattern repeating. This would then allow `UNION` to be used as a way to highlight that there is a diversity of patterns being combined.

PR #332, opened 2019-11-07
[WIP] Constrain editions without isbn

maxlath:
Editions without publisher and publication date are unmergeable has we can't know for sure which edition a user meant. Now that publishers can be created locally, adding this kind of restriction should be possible without too much disruption for the users


jum-s:
i think i understand the description of this PR, but this branch is only modifying tests, is that what its supposed to be? nothing missing ?

maxlath:
@jum-s sorry, I pushed that branch thinking it was done. I marked it as WIP

PR #336, opened 2019-11-09
transform resolve_on_labels into resolve_on_terms

maxlath:
which boils down to using the `getEntityNormalizedTerms` on both existing entities and seeds, allowing to take entities aliases into account


jum-s:
ah i was thinking of doing this exact PR today

PR #349, opened 2019-12-28
Level party update

maxlath:
`level-party` was locking us on Node v8, but that's [now solved](https://github.com/Level/party/pull/20#issuecomment-562875447) with `level-party v4.0.0`

This PR takes advantage of the update to `level-party v4.0.0` and thus to `levelup v4` to refactor some level-related logic.

Officially updating to the latest Node LTS will be done in another [branch](https://github.com/inventaire/inventaire/commits/update-node) has that's unfortunately not that trivial, due to dependencies' dependencies (namely `hyperquest`) printing undesirable warnings :/


server/db/level/utils.js
```diff
@@ -0,0 +1,35 @@
...
+      return db.batch(ops)
+    })
+  },
+
+  formatBatchOps: ops => forceArray(ops).map(setDefaultType),
+
+  // Levelup rejects `null` or `undefined` values, so the empty value should
+  // be an empty string.
+  // Useful when the key alone stores all the data that needs to be stored
+  emptyValue: ''
```
jum-s: so store an empty string in a utils file to serve it in other files like `debounce_email` below? why not using directly an empty string there instead of defining one here ?

maxlath: For the sack of documentation: it looks more explicit to me to have an `emptyValue` variable with a comment explaining why we use that, rather than just use an obscure `''` in places where it's needed. Unless you add a comment in all the places that pattern is used. But then, why not having it in one place

The same idea is used in [`lib/couch` `minKey` and `maxKey`](https://github.com/inventaire/inventaire/blob/c78e62cd/server/lib/couch.js#L26-L31).

It makes a lot of sense to me to have those values in those lib files, as those values are very much bound to respectively levelup and couchdb internals

PR #354, opened 2020-01-12
replace bluebird by native promises

maxlath:
Bluebird was useful while promises weren't mature enough in NodeJS, but now most of what Bluebird has to offer in terms of utils is either available natively or easily reproducible. The last edge would be the performance benefits that Bluebird use to claim, but even that is fading. [From Bluebird own readme](https://github.com/petkaantonov/bluebird#note):

> Promises in Node.js 10 are significantly faster than before. Bluebird still includes a lot of features like cancellation, iteration methods and warnings that native promises don't. If you are using Bluebird for performance rather than for those - please consider giving native promises a shot and running the benchmarks yourself.

Another good reason is that native promises might offer better stack traces that those:
```
TypeError: expecting an array or an iterable object but got [object Null]
    at apiRejection (/home/admin/inventaire/node_modules/bluebird/js/release/promise.js:10:27)
    at MappingPromiseArray.init (/home/admin/inventaire/node_modules/bluebird/js/release/promise_array.js:63:19)
    at Promise._settlePromise (/home/admin/inventaire/node_modules/bluebird/js/release/promise.js:566:21)
    at Promise._settlePromise0 (/home/admin/inventaire/node_modules/bluebird/js/release/promise.js:614:10)
    at Promise._settlePromises (/home/admin/inventaire/node_modules/bluebird/js/release/promise.js:693:18)
    at Async._drainQueue (/home/admin/inventaire/node_modules/bluebird/js/release/async.js:133:16)
    at Async._drainQueues (/home/admin/inventaire/node_modules/bluebird/js/release/async.js:143:10)
    at Immediate.Async.drainQueues [as _onImmediate] (/home/admin/inventaire/node_modules/bluebird/js/release/async.js:17:14)
    at processImmediate (internal/timers.js:439:21)
```

This PR focuses on making dropping Bluebird possible by
- using native promises and their methods where possible (`Promise.all`, `Promise.resolve`, `Promise.reject`)
- replacing some Bluebird-specific methods by alternative syntax (`Promise.prototype.get`)
- replacing Bluebird-specific methods by utils functions (`Promise.props`, `Promise.prototype.tap`, `Promise.prototype.filter`, `Promise.prototype.map`, `Promise.prototype.delay`)
- dropping non-essential Bluebird-specific methods that couldn't be easily replaced for now, but might need to be recovered later (`Promise.prototype.timeout`)

Other than now being able to use `async/await` (almost) anywhere in the code, this PR would also allow to take the following actions:
- replace `bluereq` by `node-fetch`, dropping the 60+ dependencies of the now deprecated `request` lib
- update `blue-cot` to v4 (where `bluereq` has been replaced by `node-fetch`)
- replace `request` by `node-fetch` while we are at it (might be hard in some places where `request` is used for its streaming capabilities though)
Bonus:
- use persistent TCP connections when it make sense, while we are at it (see [`node-fetch` doc on agents](https://github.com/node-fetch/node-fetch#custom-agent), [already implemented in `blue-cot`](https://github.com/maxlath/blue-cot/commit/43ef854))


jum-s:
nothing much to say except "shorten promises_.map to pmap, for brevity" which i disagree cuz of readability. To me, `pmap` is too short for what it does, i dont think it makes it easier to read. Also this PR contains a lot of syntax changes already, so lets take it easy on de "brievity" of stuff for now

maxlath:
@jum-s I removed the "shorten promises_.map to pmap, for brevity" commit

PR #358, opened 2020-01-24
truncate geographic positions

maxlath:
it seems that as of this day, we are guilty of keeping position coordinates at the atomic level
cf https://xkcd.com/2170/

for context (and as I wasn't sure myself), we are dealing with decimal degrees, as [that's what browsers give us](https://developer.mozilla.org/en-US/docs/Web/API/GeolocationCoordinates)

while that's not exactly a bug, relying on such precise coordinates just adds noise

but this PR also fixes a bug: it's currently impossible to update a group position, any group update returns a 400, see de4f527 and 8a043f45

### DIFF
--- a/server/controllers/entities/reverse_claims.js
+++ b/server/controllers/entities/reverse_claims.js
@@ -6,7 +6,7 @@ const sanitize = __.require('lib', 'sanitize/sanitize')
 
 const sanitization = {
   property: {},
-  value: {},
+  value: { type: 'string' },
   refresh: {
     optional: true
   },
--- a/server/controllers/entities/update_label.js
+++ b/server/controllers/entities/update_label.js
@@ -8,21 +8,23 @@ const sanitization = {
   uri: { optional: true },
   id: { optional: true },
   lang: {},
-  value: {}
+  value: { type: 'string' }
 }
 
 module.exports = (req, res) => {
   sanitize(req, res, sanitization)
   .then(params => {
     let { uri, id, value, lang } = params
+
     const prefix = getPrefix(uri, id)
     const updater = updaters[prefix]
 
-    if (uri) { id = unprefixify(uri) }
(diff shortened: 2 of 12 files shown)


jum-s:
Why not truncate even more : to 3 digits ? Your magnificent source of knowledge says "you are pointing at a specific suburbian cul-de-sac" which is the precision we want ? (and not "the corner of a house")

maxlath:
as stated in bf42bd7 commit message:

> Reluctant to remove more decimals, as, while [we discourage] most people to put their exact location their might be present or future use-cases where being able to pin-point a location would be welcome (ex: for a small public library using inventaire as their backend)

in which case, we need the 4th decimal, as, if I get it right, that would be an approximation of maximum ~20m, while truncating at the 3th decimal would be an approximation of maximum ~200m

For reference, if I get it right:
- distance between longitude 0.000 and 0.001 at the equator = (40075km/360)/1000 = 111m
- distance between longitude 0.0000 and 0.0001 at the equator = (40075km/360)/10000 = 11m
- distance between latitude 0.000 and 0.001 = (40007km/180)/1000 = 222m
- distance between latitude 0.0000 and 0.0001 = (40007km/180)/10000 = 22m

so if we consider the use case of wanting to pin-point a house, 22m can actually mean pointing the next house, in which case we would rather need to add a 5th decimal to not shoot our selves in the foot. nothing irreversible there though, so we could go with a certain decimal and totally change that later

**Bibliography**
* [gisgeography.com: Degrees/Minutes/Seconds (DMS) vs Decimal Degrees (DD)](https://gisgeography.com/decimal-degrees-dd-minutes-seconds-dms/)

jum-s:
for documentation sake : 5th digit is picked to have the maximum precise possibility (corner of house) to "not shoot our selves in the foot"

PR #364, opened 2020-02-20
documenting data transformation

maxlath:
From time to time, a transformation of database documents is needed, typically to fit a change in models code. This transformation process was handled by [`./scripts/migrator.js`](https://github.com/inventaire/inventaire/blob/35b1e63/scripts/migrator.js), which used a set of modules with names around "migration", in reference to SQL [schema migration](https://en.wikipedia.org/wiki/Schema_migration), which might have been improper given the schema-less-ish-ness of our data.

This PR proposes to replace `./scripts/migrator.js` by a set of external tools, used in a workflow documented in a new `./docs/database_administration.md` file, and to rename this operation "data transformation" rather than migration.

A major advantage of this new proposed workflow is the transparency of the process: all transformed documents can be first saved to a file for inspection, before being sent to the database.


jum-s:
works for me, at the exception of the last paragraph to verify everything `cat responses | grep conflict`, there is no such file as `responses`. Did you mean `update.errors` ?
I would also call the file `data-transformation` instead of `data-administration` to fit the purpose of the file and the name of the branch/PR

PR #379, opened 2020-04-11
Remove /api/entities?action=exists-or-create-from-seed endpoint

maxlath:
by making `/api/entities?action=resolve` take over. To reach feature parity, this PR proposes to add an `enrich` flag to the endpoint, which would allow to add data from dataseed to the resolve entries. In the current state, that would only mean to look for a book cover corresponding to a given ISBN on dataseed

This PR is coupled to https://github.com/inventaire/inventaire-client/pull/212


server/controllers/entities/lib/resolver/create_entity_from_seed.js
```diff
@@ -46,6 +47,11 @@ const createEdition = async (edition, works, userId, batchId) => {
   // garantee that an edition shall not have label
   edition.labels = {}
 
+  if (enrich === true) {
+    const { url: imageHash } = await dataseed.getImageByIsbn(isbn)
+    if (imageHash) claims['invp:P2'] = [ imageHash ]
```
jum-s, line 52: this override any seed images provided, what about the case where images are provided in the seed, but not all of them, so we would like a soft update (only create if `invp:P2` doesnt exists) ?

maxlath: yes. but if an image was provided, it would probably rather be as a URL than an image hash, so we also need to add a step to allow conversion

server/controllers/entities/lib/resolver/create_entity_from_seed.js
```diff
@@ -71,6 +77,8 @@ const createEntityFromSeed = params => {
   .then(addCreatedUriToSeed(seed))
 }
 
+// Difference between seedClaims and entityClaims:
+// seedClaims come from the request input, and might not be valid
```
jum-s, line 81: The difference is less the data quality than the data origin : seed being from an external source, while entity is the current inventaire entity

server/controllers/entities/lib/resolver/create_entity_from_seed.js
```diff
@@ -79,24 +87,30 @@ const buildClaims = (seedClaims, entityClaims, type) => {
   return entityClaims
 }
 
-const addCreatedUriToSeed = entryEntity => createdEntity => {
+const addCreatedUriToSeed = seed => createdEntity => {
```
jum-s, line 90: nice fix, an `entryEntity` is indeed a `seed`

server/controllers/entities/lib/resolver/create_entity_from_seed.js
maxlath: done in 396c917d and bbeb3235

server/controllers/entities/lib/resolver/create_entity_from_seed.js
maxlath: `entityClaims` is not the claims of a currently existing entity, but the `claims` object created higher in the file, and being enriched. I refactored (29752066) to try to make that clearer

jum-s (approved):

maxlath:
rebased, but wait for client branch to be ready to merge (to keep master deployable)

jum-s:
To me, this indeed achieve parity of features with `exists-or-create-from-seed`
TODO : use enrich to feed publishers labels to the resolver

PR #381, opened 2020-04-12
add rich csv inventory export

maxlath:
a first implementation of a csv export of a user inventory, that includes data from items entities graph


jum-s:
merge?

PR #386, opened 2020-04-27
resolver: allow to force update specific claims

jum-s:
Replace entity claims with entry claims of properties declared in the `forceUpdateProps` parameter (renaming welcome)

Usecase: an editor catalogue, representing the most canonical data is passed through the resolver. But some formating of this catalogue may be less precise than existing inv entities claims.

`forceUpdateProps` allow to update only the publication date for example, date format being less subjected to interpretation than a title.

- early parameter validation: against properties values constraints list
- perform an update, aka only if claim already exists (allowing to not check entity type before update)
- rely on not updating wikidata items policy already implemented in
update_resolved_entry

### DIFF
--- a/server/controllers/entities/lib/resolver/update_resolved_entry.js
+++ b/server/controllers/entities/lib/resolver/update_resolved_entry.js
@@ -3,27 +3,27 @@ const __ = CONFIG.universalPath
 const _ = __.require('builders', 'utils')
 const entities_ = require('../entities')
 
-module.exports = (userId, batchId) => entry => {
+module.exports = params => entry => {
+  const { reqUserId: userId, batchId, forceUpdateProps } = params
   const { edition, works, authors } = entry
 
   const allResolvedSeeds = [ edition ].concat(works, authors).filter(hasUri)
 
-  return Promise.all(allResolvedSeeds.map(updateEntityFromSeed(userId, batchId)))
+  return Promise.all(allResolvedSeeds.map(updateEntityFromSeed(userId, batchId, forceUpdateProps)))
   .then(() => entry)
 }
 
 const hasUri = seed => seed.uri != null
 
-const updateEntityFromSeed = (userId, batchId) => seed => {
+const updateEntityFromSeed = (userId, batchId, forceUpdateProps) => seed => {
   const { uri, claims: seedClaims } = seed
   if (!uri) return
 
   const [ prefix, entityId ] = uri.split(':')
   // Do not try to update Wikidata for the moment
   if (prefix === 'wd') return
-
-  return getEntity(prefix, entityId)
-  .then(addMissingClaims(seedClaims, userId, batchId))
+  getEntity(prefix, entityId)
+  .then(updateClaims(seedClaims, userId, batchId, forceUpdateProps))
 }
 
 const getEntity = (prefix, entityId) => {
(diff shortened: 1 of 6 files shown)


server/controllers/entities/lib/resolver/update_resolved_entry.js
```diff
@@ -34,10 +34,29 @@ const getEntity = (prefix, entityId) => {
...
+  if (!_.isEqual(claims, currentDoc.claims)) {
+    entities_.putUpdate({ userId, currentDoc, updatedDoc })
+  }
+}
+
+const addMissinClaims = (seedClaims, userId, batchId, entity) => {
   const newClaims = _.omit(seedClaims, Object.keys(entity.claims))
   if (_.isEmpty(newClaims)) return
-  return entities_.addClaims(userId, newClaims, entity, batchId)
+  entities_.addClaims(userId, newClaims, entity, batchId)
```
maxlath, line 61: same: return promises

tests/api/entities/resolver/resolve_and_force_update.test.js
```diff
@@ -0,0 +1,87 @@
...
+  })
+
+  it('should replace edition claim values when properties to force is specified', async () => {
+    const editionEntry = buildEditionEntry()
+    const editionUri = `isbn:${editionEntry.edition.isbn}`
+    await ensureEditionExists(editionUri)
+
+    const propertiesToForce = [ 'wdt:P123' ]
+    await resolveAndForceUpdate(editionEntry, propertiesToForce)
+    .then(Wait(100))
```
maxlath, line 54:
```suggestion
    await wait(100)
```

server/controllers/entities/lib/resolver/update_resolved_entry.js
```diff
@@ -34,10 +34,29 @@ const getEntity = (prefix, entityId) => {
...
+const forceUpdateClaims = (seedClaims, userId, batchId, forceUpdateProps, currentDoc) => {
+  const updatedDoc = _.cloneDeep(currentDoc)
+  const { claims } = updatedDoc
+  forceUpdateProps.forEach(prop => {
+    if (claims && claims[prop]) {
+      claims[prop] = seedClaims[prop]
+    }
+  })
+  if (!_.isEqual(claims, currentDoc.claims)) {
+    entities_.putUpdate({ userId, currentDoc, updatedDoc })
```
maxlath, line 54: by calling `entities_.putUpdate` directly, it skips `validateClaims` (`validateAndFormatClaims` on current master branch): that's a problem, as mistakes happens, and this validation can be precious to avoid a wide range of errors (ex: setting an entity of type `work` in place of a `publisher` or setting an edition ISBN to an ISBN already used by another edition)

maxlath (changes requested):

maxlath:
adding only the missing values was the low hanging fruit, trying to update existing data en mass is much more challenging: I would be in favour of postponing this PR to when we can refactor it with the fruits of #400

jum-s:
yep, replacing uris looks challenging indeed. But some "hardcoded" values (as non uris values) could be automatically updated, this is exactly what #507 is trying to do.
It can be discussed in a new issue, but if "a wd entity replaces an inv entity as publisher"  what should be done is to update the claim but also to try to "garbage collect" "orphan" publishers by redirecting them to the wd entity.
I dont think there is any reason to refuse en mass claim updates that can be updated individually, as long as tool to reverse it exists (that's why batch_id exists).

PR #389, opened 2020-05-10
replace bluereq and request by node-fetch

maxlath:
[`bluereq`](https://github.com/maxlath/bluereq) meant:
- mixing `bluebird` promises and non-bluebird promises, opening concerns of incompatibility (which might be unfounded, but still existing)
- depending on the now deprecated `request` module, and all its dependencies
- more maintenance work

[`node-fetch`](https://github.com/node-fetch/node-fetch):
- designed for promises
- small but powerful:
  - support streaming
  - support custom agent
- while [its future in NodeJS core is nothing certain](https://github.com/nodejs/node/issues/19393), it's built on the [standardized fetch API](https://fetch.spec.whatwg.org/), which means that what we learn about fetch in the server side is mostly usable in the client-side with browsers' fetch implementations

The transition wasn't that straight-forward given the deep coupling with bluereq features, especially:
- content-type management (to parse or not to parse the response as JSON)
- oauth: was fully handled by `request`, while you're let alone with `node-fetch`. Fortunately, most of [that transition had been already done for wikibase-edit](https://github.com/maxlath/wikibase-edit/blob/f4a2901/lib/request/oauth.js)
- reproducing the error format

This PR also comes with some extras, which are not strictly matching the PR title:
- some async/await refactor
- some code cleanup, such as drastically reducing the work done by `getThumData`


server/controllers/auth/wikidata_oauth.js
```diff
@@ -3,8 +3,26 @@ const __ = CONFIG.universalPath
 const requests_ = __.require('lib', 'requests')
 const error_ = __.require('lib', 'error/error')
 const root = CONFIG.fullPublicHost()
+const OAuth = require('oauth-1.0a')
+const crypto = require('crypto')
+const createHmacSha1Hash = (baseString, key) => {
+  return crypto.createHmac('sha1', key)
```
jum-s, line 9: consider `sha256` => https://github.com/ddo/oauth-1.0a#nodejs

server/data/openlibrary/check_cover_existance.js
```diff
@@ -2,17 +2,12 @@ const __ = require('config').universalPath
...
-const checkHeader = (url, res) => {
-  const contentType = res.headers['content-type']
-  // Coupled with OpenLibrary response headers
-  if (contentType && contentType === 'image/jpeg') {
-    return url
-  } else {
+module.exports = async (url, expectedContentType) => {
+  const { headers } = await requests_.head(url)
+  const contentType = headers['content-type']
+  if (expectedContentType != null && contentType !== expectedContentType) {
```
jum-s, line 8: might be more readable to actually [pass a flag `?default=false`](https://openlibrary.org/dev/docs/api/covers) on the request to OL. and expect a 404 when no image available

server/controllers/auth/wikidata_oauth.js
maxlath: unfortunately, when trying with `sha256`, Wikidata answers with `Error: An error occurred in the OAuth protocol: Signature method 'HMAC-SHA256' not supported try one of the following: HMAC-SHA1, RSA-SHA1`

server/data/openlibrary/check_cover_existance.js
maxlath: done in 79681beb

PR #401, opened 2020-05-24
Move entity to wd

jum-s:
a light action which return aliases hardcoded in `server/lib/wikidata/aliases.js`
input parameters: lang, type
output keys: uri, id and label per lang

to be merged before client branch `move-entity-to-wd`


server/controllers/entities/lib/move_to_wikidata.js
```diff
@@ -18,6 +18,7 @@ module.exports = (user, invEntityUri) => {
   })
   .then(entity => {
     const { labels, claims } = entity
+    rewriteP31(claims, asP31value)
```
jum-s, line 21: on the fly rewriting before pushing, allowed only because of entity destruction after pushing action, otherwise would require an entity update before pushing

jum-s:
closing this, to be continued on #416

Issue #295, opened 2019-04-30
allow different types as possible values for some claims


maxlath:
### P50
It is today impossible to add anything that isn't an entity identified as being of the type `human` to an 'author' (`P50`) claim, while there are many books for which no single author can be identified.

Example: works published by collectives where the individuals aren't identified or change at every edition (ex: Le Comité Invisible, Lonely Planet, UNICEF publications, etc)

### P123
the publisher (`P123`) of an edition might be an entity of type `human` ([Wikidata has 6000 of those](https://query.wikidata.org/#SELECT%20%28COUNT%28%3Fitem%29%20AS%20%3Fcount%29%20%7B%0A%20%20%3Fitem%20wdt%3AP123%20%3Fhuman%20.%0A%20%20%3Fhuman%20wdt%3AP31%20wd%3AQ5%20.%0A%7D%0A))

### P127
A comparable problem is encountered for 'owned by' (`P127`) claims, for which values might be both humans and organizations

## Proposed solution
* some claims should allow different entity types
* we need a new entity type that could be called `organizations`, and that would be a superset of publishers (which should also generally be accepted as organizations, but with the particularity of publishing books), and cover the needs for both companies that may own publishers and collectives that may author some works

## Doubts
* Maybe publishers should just be fully integrated to this new `organizations` type? The problem would then be that the client wouldn't know if it needs to fetch the works from that possible P50 value or the editions of that possible P123 value. But maybe it could fetch both?

maxlath:
* [x] some claims should allow different entity types: done with 9073d33db3ba4f5b8a84f2bab37516e11ea05df5 which  pluralize and rename `restrictedType` to `entityValueTypes`

PR #407, opened 2020-06-24
add support for local publisher collections entities

maxlath:
* define the list of accepted properties for the type 'collection'
* add a `/api/entities?action=publisher-publications&uri=${publisher-uri}` endpoint to ease the display of publishers collections and editions on publisher's layouts

somewhat loosely connected to the title of the PR, but added for free:
* tighten-up constraints on editions to limit zombie editions with no ISBN, no publisher, no publication date: possible now that its allowed to create local publishers
* refactor entity validation to have a central place that validates controlled properties (such as those zombie editions mentioned above) on each operations on entity docs

associated with client PR https://github.com/inventaire/inventaire-client/pull/211


server/controllers/entities/lib/properties/properties_values_constraints.js
```diff
@@ -37,6 +37,8 @@ module.exports = {
   'wdt:P31': extend(bases.uniqueEntity, { adminUpdateOnly: true }),
   // author
   'wdt:P50': extend(bases.humanEntity, { hasPlaceholders: true }),
+  // editor
+  'wdt:P98': bases.humanEntity,
```
jum-s, line 41: It seems Wikiproject:Books recommend a broader definition of P98 values than only humans. If you take a look at the P98 [example for a work](https://www.wikidata.org/wiki/Wikidata:WikiProject_Books) (as nothing of the matter is recommended for collections) the example shows the statement `Q25666 P98 Q299015`.

[15% of entities which belong to a P98 are not Q5s](https://query.wikidata.org/#select%20%3Feditor%20%3FeditorLabel%20%3Fcollection%20%3FcollectionLabel%20%7B%0A%20%20%3Fcollection%20wdt%3AP98%20%3Feditor.%0A%20%20filter%20not%20exists%20%7B%3Feditor%20wdt%3AP31%20wd%3AQ5%7D%0A%20%20SERVICE%20wikibase%3Alabel%20%7B%20bd%3AserviceParam%20wikibase%3Alanguage%20%22%5BAUTO_LANGUAGE%5D%2Cen%22.%20%7D%0A%7D) (3513/23794) and I believe a best effort cannot avoid 15% (and you may remember this sentence future self, its written to eternity now) 

[Among those, many looks like organisations](https://query.wikidata.org/#select%20%3FinstanceLabel%20%3Finstance%20%28count%28%3Finstance%29As%20%3Fcount%29%20%7B%0A%20%20%3Fcollection%20wdt%3AP98%20%3Feditor.%0A%20%20filter%20not%20exists%20%7B%3Feditor%20wdt%3AP31%20wd%3AQ5%7D%0A%20%20%3Feditor%20wdt%3AP31%20%3Finstance%0A%20%20SERVICE%20wikibase%3Alabel%20%7B%20bd%3AserviceParam%20wikibase%3Alanguage%20%22%5BAUTO_LANGUAGE%5D%2Cen%22.%20%7D%0A%7D%0Agroup%20by%20%3Finstance%20%3FinstanceLabel%0Aorder%20by%20desc%28%3Fcount%29). 

For example, can we really avoid editors such as Q202660 (Collège de France) ?

Since adding P98 to inventaire is a creation, i think it could be nice to also tackle this " `organisations` as full fledged citizen in inventaire data model", as partly discussed in #295 but for other props.

tests/api/entities/editions/create.test.js
```diff
@@ -36,6 +38,43 @@ describe('entities:editions:create', () => {
     .catch(done)
   })
 
+  it('should reject an edition without an ISBN without a publisher', async () => {
```
jum-s, line 41: if i understand this well, it restricts quite a lot edition creation, isnt it ? What if my old book's publisher is unknown to me ?

maxlath: @jum-s Are there books where the publisher is unknown? how else can we prevent the creation of  uneditable, unmergeable editions not connected to anything?

maxlath: a compromise could be to let it unrestricted on the server side but to request that the client pushes the user to avoid creating isolated editions, but that would mean that each client will have to re-implement this limitation, and that anyone directly hitting the API will be free to create isolated/junk editions

([example of an isolated edition that can't be edited or merged because we don't have an ISBN or the year of publication](https://inventaire.io/entity/inv:0888c0e93595f51c17f601eb6a85fed4))

server/controllers/entities/lib/properties/properties_values_constraints.js
maxlath: @jum-s I agree on the principle, this let us with 4 options:
- let the current implementation, which is too restrictive: could be an easy first step
- remove restriction and accept any type of entity until we can more finely restrict input: would be equally trivial
- create an organization type and implement the possibility to have properties bound to several types (in this case humans+organizations): while that's what we should probably do on the long term, I would oppose picking this option as a premise to merging this PR as that's a guaranteed staled branch imho
- remove all references to P98 from this PR: trivial

tests/api/entities/editions/create.test.js
maxlath: I removed the constrain in fa2cc519 to not block this PR on this

server/controllers/entities/lib/properties/properties_values_constraints.js
jum-s: ouch 3 months to answer this... option 2 seems attractive, we need to consider what to do with values which would be created while implementing options 3 and which would not fit `humans` or `collectivity` (yes im planting seed for the naming)

jum-s: as `collection P98 publisher` is hardly used on Wikidata and because [wd community](https://www.wikidata.org/wiki/Property_talk:P98) tends to define P98 as a property dedicated to academic or scientific newspapers, it appeared that `P98` integration in an entity edit form can be on hold for the moment as it might add some confusion between P98 and P123, until a consensus for its usage will be found.

jum-s (approved):

PR #412, opened 2020-08-10
replace wd:Q571 by wd:Q47461344 as default work P31

maxlath:
Q571 as a P31 value has been deprecated for a while on Wikidata, meaning that [Wikidata items created from an inventaire.io import need to be fixed](https://editgroups.toolforge.org/b/wikibase-cli/cbcf544f2f519/). This patch fixes the problem at the source.


jum-s:
this patch fixes the creation of a work as a Q47461344. Shall the actual Q571 works (300k+) created a while ago also be moved to Q47461344 ? 
[wikiproject Books](https://www.wikidata.org/wiki/Wikidata_talk:WikiProject_Books/2019#Idea_of_cleanup) seems to not have a consensus on this question (tldr. its because of the editions situation, which inventaire is not affected by), so i would also consider this massive entities bulk update

PR #424, opened 2020-09-14
temporarily stop to make requests to a host that answered with a timeout

maxlath:
Among the SPARQL endpoints the resolver makes requests to, there is one that currently refuses to answer anything, which always end-up with a timeout after 60 seconds, which makes eeeeeeeverything slow. This PR brings a mechanism to automatically identify hosts that fail to answer and temporarily ban requests to those, with a growing ban time


server/lib/requests.js
```diff
@@ -18,6 +20,9 @@ const req = method => async (url, options = {}) => {
   assert_.string(url)
   assert_.object(options)
 
+  const { host } = new URL(url)
+  checkHostBan(host)
```
jum-s, line 24:
```suggestion
  throwIfBanned(host)
```

PR #432, opened 2020-09-21
Shelves

jum-s:
main user cannot create or update an item with shelves belonging to someone else


jum-s:
that should be mergeable now

maxlath:
merged and running in prod :tada:

Issue #361, opened 2020-02-13
paginate notifications


maxlath:
At the moment, [the server only allows clients to fetches all the notifications at once](https://github.com/inventaire/inventaire/blob/878cdca9/server/controllers/notifications.js#L8-L14) (via `GET /api/notifications`). That's fine for now as the volume isn't enormous (my own account, which is probably among the "large" ones, has only 30 notifications, which represent 8 ko of data), but as it's fetched every time the client is opened, it could be a nice future-proofing task to optimize it a bit:
- allow to only fetch a count of unread notifications
- pagination: allow to only fetch the 10 last notifications, and then the 10 next, etc

maxlath:
fixed by #441

PR #441, opened 2020-10-31
paginate notifications

Aethelflaed:
I tried my hand at paginating the notifications as requested in #361

This is my first time working with this stack so I may have missed something. I used code similar to what I've found for the other APIs.

I found a corresponding test file, but I haven't found any way of actually running it (yet).

Comments and pointers welcome so I can improve :)

### DIFF
--- a/server/controllers/notifications.js
+++ b/server/controllers/notifications.js
@@ -1,14 +1,42 @@
 const __ = require('config').universalPath
 const _ = __.require('builders', 'utils')
 const error_ = __.require('lib', 'error/error')
+const sanitize = __.require('lib', 'sanitize/sanitize')
 const responses_ = __.require('lib', 'responses')
 const notifs_ = __.require('lib', 'notifications')
+
+const sanitization = {
+  limit: { optional: true, default: 10 },
+  offset: { optional: true },
+}
+
+const Paginate = page => notifications => {
+  let { limit, offset } = page
+  const total = notifications.length
+  if (offset == null) offset = 0
+  const last = offset + limit
+
+  if (limit != null) {
+    notifications = notifications.slice(offset, last)
+    const data = { notifications, total, offset }
+    if (last < total) data.continue = last
+    return data
+  } else {
+    return { notifications, total, offset }
+  }
+}
+
+const getNotifications = page => {
+  return notifs_.byUserId(page.reqUserId)
+  .then(Paginate(page))
+}
 
 const get = (req, res) => {
   if (req.user == null) return error_.unauthorizedApiAccess(req, res)
 
-  return notifs_.byUserId(req.user._id)
-  .then(responses_.Wrap(res, 'notifications'))
+  return sanitize(req, res, sanitization)
+  .then(getNotifications)
+  .then(responses_.Send(res))
   .catch(error_.Handler(req, res))
 }
 
--- a/server/controllers/notifications.js
+++ b/server/controllers/notifications.js
@@ -10,8 +10,8 @@ const sanitization = {
   offset: { optional: true },
 }
 
-const Paginate = page => notifications => {
-  let { limit, offset } = page
+const paginate = params => notifications => {
+  let { limit, offset } = params
   const total = notifications.length
   if (offset == null) offset = 0
   const last = offset + limit
@@ -26,9 +26,9 @@ const Paginate = page => notifications => {
   }
 }
 
-const getNotifications = page => {
-  return notifs_.byUserId(page.reqUserId)
-  .then(Paginate(page))
+const getNotifications = params => {
+  return notifs_.byUserId(params.reqUserId)
+  .then(paginate(params))
 }
 
 const get = (req, res) => {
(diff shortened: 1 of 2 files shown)


maxlath:
you can run API tests with
```sh
npm run test-api-quick ./tests/api/notifications/get.test.js
```
This will start a server on port 3009 with dedicated databases (using the suffix `-tests`), and then run the test file.

I pushed a commit (9a5c3560) on master to update the existing test to the new recommended (and hopefully simpler) async/await syntax.

I pushed another commit (7113683d) on master that should save you the need to setup the [`entities-search-engine` server](https://github.com/inventaire/entities-search-engine): you don't need it to work on notifications, and there is ongoing work to reintegrate it to the main server, so no need for you to bang your head on that one ;)

Aethelflaed:
Okay, I managed to get the tests started and added one to check for total presence & type, is that good enough or am I missing something?

server/controllers/notifications.js
```diff
@@ -1,14 +1,42 @@
...
+const sanitize = __.require('lib', 'sanitize/sanitize')
 const responses_ = __.require('lib', 'responses')
 const notifs_ = __.require('lib', 'notifications')
 
+const sanitization = {
+  limit: { optional: true, default: 10 },
+  offset: { optional: true },
+}
+
+const Paginate = page => notifications => {
```
jum-s, line 13:
```suggestion
const paginate = params => notifications => {
```
Thanks for posting this because we can now update the user contribution guidelines (after https://github.com/inventaire/jingo/issues/6) with this kind of naming conventions that are not really documented, 
The suggestion above is welcome but not mandatory, it may be merged like this in a few days if you dont feel like modifying it.

tests/api/notifications/get.test.js
```diff
@@ -5,7 +5,8 @@ const { authReq } = __.require('apiTests', 'utils/utils')
 
 describe('notifications:get', () => {
   it('should get user notifications', async () => {
-    const { notifications } = await authReq('get', '/api/notifications')
+    const { notifications, total } = await authReq('get', '/api/notifications')
```
jum-s, line 8: sooo, lets say this part of the code is far from having a good test coverage :) ... Since some commits are done, it would be great to detail a bit the tests to check what is a notification (ie. its type). But I dont think its necessary to go as far as testing the pagination itself.

a notification is triggered when making new friends: 

```js
const { makeFriends } = __.require('apiTests', 'utils/relations')
...
await makeFriends(user, friend)
```

again same comment as above this suggestion is optional

Aethelflaed: I kept the `total` assertion to ensure the pagination change has taken effect and added a notification as you said but I have seemingly random failure on this test (like one failure after running the test 30 times)

I got "lucky" the first time I guess, then started loop-testing the function.

Is it possible that the `makeFriends`' notification is not directly available for reading? Should I add a `sleep(1)`?

jum-s: i cant reproduce the failure.
Indeed you may want to introduce a "sleep" with the `wait` function in `lib/promises`, even though i dont see why would be necessary.

tests/api/notifications/get.test.js
```diff
@@ -1,12 +1,22 @@
...
+    const user = await getUser()
+    const friend = await getReservedUser()
+    await makeFriends(user, friend)
+
     const { notifications, total } = await authReq('get', '/api/notifications')
     notifications.should.be.an.Array()
-    total.should.be.aboveOrEqual(0)
+    total.should.be.equal(1)
+
+    const notif = notifications[0]
```
jum-s, line 17: the above comment impact this line too, you may want to pop the array to be sure to pick the lasted created notif

tests/api/notifications/get.test.js
maxlath: short version: adding a 1s delay should work in most cases (that is, when the server isn't under heavy fire?). Or wait for #446 to be merged.

longer version: when accepting the friend request, a `notify:friend:request:accepted` event is triggered, letting the [notifications event listener trigger the `acceptedRequest` callback](https://github.com/inventaire/inventaire/blob/874b095/server/lib/notifications.js#L112), but the server doesn't wait for that async callback to be done to answer, so when waiting for `await makeFriends`, you have no guaranty that the notification was already posted on the notifications database... 

Those event listeners are not very developer-friendly for tests, but they express the idea that notifications/emails/tracking events are not essential: if the server fails to do one of those, that shouldn't make the HTTP request return an error. But, in tests, having that guaranty would be preferable, so here is a fix proposition https://github.com/inventaire/inventaire/pull/446

maxlath: btw, to reliably reproduce the failure (and thus test the #446 patch), you can turn that `acceptedRequest` function into an `async` function and add `await wait(5000)`

Aethelflaed: I was suspecting something like that, and I'll wait for the patch then. Thanks for the explanation !

maxlath: I just merged #446 so rebasing on master should solve your issue

Aethelflaed: done

Issue #431, opened 2020-09-21
add shelves data to csv export


maxlath:

Aethelflaed:
This could be closed I think?

maxlath:
yep! :tada:

PR #443, opened 2020-11-02
add shelves data to csv export

Aethelflaed:
Implementation proposal for #431


jum-s:
:tada: nice! its now in production

Issue #430, opened 2020-09-21
allow to get a list of items that are in no shelves


maxlath:
that would allow the client
* to display a "no shelf" shelf: addressed by #445
* to suggest those books first in the "add books to shelf" menu

maxlath:
the server side part being done #445, remains the client side: I would suggest to wait for the Svelte refactor of the inventory browser, as adding a "no shelf" element in a list in Backbone.Marionette requires some ugly hack (namely adding a pseudo model to the collection), while that should be much easier in Svelte

jum-s:
client side is done with svelte rewrite

PR #445, opened 2020-11-03
add user inventory-view for items without shelf

Aethelflaed:
proposal for #430


maxlath (changes requested):
Nice! I'm a bit unsure about the `get_by_access_level_without_shelf`, so in this review, I mostly try to come with ~alternative implementations~ [an alternative implementation] to save the duplicated code, ~but I'm not sure which implementation to prefer, opinions welcome~

maxlath:
@Aethelflaed let me know if you would like more time to work on this PR, or if I should take over: either is fine with me :)

Aethelflaed:
Yes please :) I'd like to try a few things based on your comments

server/controllers/items/inventory_view.js
```diff
@@ -12,7 +12,8 @@ const shelves_ = __.require('controllers', 'shelves/lib/shelves')
 const sanitization = {
   user: { optional: true },
   group: { optional: true },
-  shelf: { optional: true }
+  shelf: { optional: true },
+  noShelf: { optional: true, generic: 'boolean' }
```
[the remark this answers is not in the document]
Aethelflaed: I just learned how to rename attributes while destructuring and done!

tests/api/items/inventory_view.test.js
```diff
@@ -31,6 +32,22 @@ describe('items:inventory-view', () => {
     res.itemsByDate.should.be.an.Array()
   })
 
+  it('should return an inventory-view for user items without shelf', async () => {
+    const user = await createUserWithItems()
+    const shelf = await createShelf(user)
+    let res = await publicReq('get', `${endpoint}&user=${user._id}&no-shelf=true`)
```
Aethelflaed, line 38: As far as I can see, `createUserWithItems` does NOT create the indicated number of items (10), I haven't figured out why yet, but it sometimes create only one item which causes the `shelves?action=add-items` to fail.

Any idea on how to proceed on this point?

db/couchdb/design_docs/items.json
```diff
@@ -25,6 +25,9 @@
     },
     "byShelvesAndListing": {
       "map": "(doc)->\n  if doc.shelves?\n    for shelf in doc.shelves\n      emit [shelf, doc.listing], null"
+    },
+    "byListingWithoutShelf": {
+      "map": "(doc)->\n  if ! doc.shelves?\n    emit [doc.owner,doc.listing], null"
```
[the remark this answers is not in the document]
Aethelflaed: From what I've seen, this would work because the IDs are unique even between collection (or whatever it is called in couchdb) but it's still a "dirty hack" in a way, as the name can't let you predict what it would do and the answer would still look like the expected result.
Using this modified request would mean changing all the `build_keys_from_shelf`, or the `byShelf`/`byShelves` code, further obfuscating the intent.

I understand the need to reduce code duplication, but in this case I'd suggest to propagate the `noShelf` parameter to `getAuthorizedItems.byUser` then to `getByAccessLevel`.

What do you think?

tests/api/items/inventory_view.test.js
maxlath: @Aethelflaed that's weird, do you see any error in the test server logs? (fyi, you can follow those logs with `tail -f ./logs/test-server.log -n 100`)

I added a debug commit to https://github.com/inventaire/inventaire/tree/430-get-user-items-without-shelf

db/couchdb/design_docs/items.json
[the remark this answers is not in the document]
maxlath: It could be made more explicit that this is a pseudo shelf id based on a user id, example:
```js
const defaultUserShelf = { _id: `u:${userId}:default`, owner: userId }
```
and that wouldn't change anything in `build_keys_from_shelf` as far as I can tell. What I like about this solution is that it underlines that it's a shelf concern. We could even push the logic to replace the new `no-shelf` parameter by a user-id based pseudo-shelf-id that would then carry that role through the whole stack, from the client API call to the database view, and could have other yet-unknown applications.

But it does introduce a new concept (pseudo-shelf-ids), which would require to update validation, and might create some weirdness here and there, so if the solution you suggest to propagate the `noShelf` parameter works and is less weird, we can go with that :)

On another subject but still related to those CouchDB views, beware that you need to check for the case where `doc.shelves= []`, which should be considered to be a 'no shelf' case
```suggestion
    "byShelvesAndListing": {
      "map": "(doc)->\n  if doc.shelves?.length > 0 \n    for shelf in doc.shelves\n      emit [shelf, doc.listing], null"
    },
    "byListingWithoutShelf": {
      "map": "(doc)->\n  if (not doc.shelves?) or doc.shelves.length is 0\n    emit [doc.owner,doc.listing], null"
```

tests/api/items/inventory_view.test.js
Aethelflaed: @maxlath thanks, I don't know why I didn't check the server logs, the answer is quite easy indeed: we (as in my code but the previous inventory-view test also) make a public request on the inventory view but there's no guarantee that there's going to be any public items.

I slighlty changed createUserWithItems to ensure at least two public items

db/couchdb/design_docs/items.json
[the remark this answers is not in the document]
Aethelflaed: I made the change for `doc.shelves.length is 0` and propagated the `withoutShelf` argument (I renamed it for consistency with the view)

server/controllers/items/lib/get_authorized_items.js
```diff
@@ -8,9 +8,9 @@ const getInventoryAccessLevel = require('./get_inventory_access_level')
 
 // Return what the reqUserId user is allowed to see
 module.exports = {
-  byUser: async (userId, reqUserId) => {
+  byUser: async (userId, reqUserId, opts = {}) => {
     const accessLevel = await getInventoryAccessLevel(userId, reqUserId)
-    return getByAccessLevel[accessLevel](userId)
+    return getByAccessLevel[accessLevel](userId, reqUserId, opts)
```
Aethelflaed, line 13: I had to specify `reqUserId` again here, I guess it's not a problem?

tests/api/fixtures/populate.js
```diff
@@ -15,9 +15,9 @@ const API = module.exports = {
     return populatePromise
   },
 
-  createUserWithItems: async username => {
```
Aethelflaed, line 18: I checked the code base and the only instance where this argument is used, it's an object, and an object is expected in `createUser` too, so I changed the name

tests/api/items/inventory_view.test.js
```diff
@@ -31,6 +33,23 @@ describe('items:inventory-view', () => {
     res.itemsByDate.should.be.an.Array()
   })
 
+  it('should return an inventory-view for user items without shelf', async () => {
+    const user = await createUserWithItems({}, _.times(2, () => Object.new({ listing: 'public' })))
```
maxlath, line 37: The test fails because this line throws `TypeError: Object.new is not a function`

```suggestion
    const user = await createUserWithItems({}, _.times(2, () => ({ listing: 'public' })))
```
Explaination: "[In an arrow function] To return an object literal expression requires parentheses around expression" [[source](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions#Advanced_syntax)]

or alternatively

```suggestion
    const itemsData = [ { listing: 'public' }, { listing: 'public' } ]
    const user = await createUserWithItems({}, itemsData)
```

Once fixed, the test passes

maxlath (changes requested):

tests/api/items/inventory_view.test.js
Aethelflaed: done!

maxlath (approved):

maxlath:
nice! :tada:

Issue #84, opened 2017-03-05
add OAuth provider


maxlath:
Would offering OAuth authentification be of any use?

Using this issue to centralize documentation on the subject:
- [Creating an API authenticated with OAuth 2 in Node.js](http://resources.infosecinstitute.com/securing-web-apis-part-ii-creating-an-api-authenticated-with-oauth-2-in-node-js/)
- [The Most Common OAuth2 Vulnerability ](http://homakov.blogspot.fr/2012/07/saferweb-most-common-oauth2.html)
- [node-oauth2-server](https://github.com/oauthjs/node-oauth2-server)

jum-s:
Works with already integrated passportJS lib, some documentation :

- https://www.ibm.com/developerworks/library/wa-mean5/index.html
- https://aaronparecki.com/oauth-2-simplified/
- http://passportjs.org/docs/oauth

Next step : Choose a provider, Im not sure i can be in favor a Facebook... Github seems to far from the users 

Then I could give it a try, since I implemented on a similar app.

Since inventaire already implement "express-session", and passport local strategy, it **should** be seamless to integrate an oauth2 provider

maxlath:
sorry, the issue title and description might not be explicit enough: this isn't about using external providers to authentify users, but to make inventaire one of those providers so that external application can interact with users inventories on their behalf without having to ask for their username and password, and while being limited to only the rights they are granted

jum-s:
then a question (with the answer in it) : is it the objective of inventaire to be a oauth provider?

maxlath:
@jum-s I updated the first comment in this issue with a few use cases that do seem like valid reason to provide oauth

pandark:
Having the user's credentials is not a viable option if you want to (for example)…
* host book listing (with more infos, for example) for people other than yourself (family, friends, clients…) and push it to inventaire.io
* publish an online import tool for a format not (yet) managed by inventaire.io, using the API
* …

maxlath:
resources:
* [How to Hack OAuth • Aaron Parecki](https://www.youtube.com/watch?v=tbu4CfzP25o)

maxlath:
Now that #487 was merged, the basic OAuth server infrastructure is there, and used in prod for authentication on https://wiki.inventaire.io . What's missing for third party apps to use it:
* [ ] add new scopes that could be useful to third party apps ([current scopes](https://github.com/inventaire/inventaire/blob/master/server/controllers/auth/lib/oauth/scopes.js))
[...]

PR #458, opened 2020-11-17
move to shorter-lived and safer cookies

maxlath:
Currently, cookies are issued for 10 years, which is very convenient for users who don't like to have to login (typically because of a bad configuration or absence of a password manager), and to API users who then may retrieve long life cookies and use it as authentication tokens for the 10 next years, but that's not really best practices.

The current PR proposes to:
- move to 6 month cookies by default, after which cookies will be expire in the browser (`maxAge`) and be rejected by the server if presented by some other mean (`enforceSessionMaxAge`). Consequently, GUI users will have to re-login at least once every 6 months, and API users will have to update their cookies (until we give a better option with OAuth #84)
- update cookie-session options:
    - use a custom name: so `express` and `express.sig` cookies become `inventaire` and `inventaire.sig` by default
    - replace `secret` by `keys` with an instance of [Keygrip](https://github.com/crypto-utils/keygrip) (as [allowed by cookie-session](https://github.com/expressjs/cookie-session/#keys), to be able to replace `sha1` by `sha256` as a signature algorithm
    - set `sameSite` as we only use first-party context
    - expliciting the default value, for the sake of documentation but also to cover for the somewhat improbable case where the default would change

Following actions:
- as cookie sessions will be invalidated anyway by the new signature algorithm, take the opportunity to change the secret
- in the future, it could be nice to have a signature key rotation mechanism, which is made quite simple by [Keygrip](https://github.com/crypto-utils/keygrip)

Some readings:
* [cookie-session: deleting the session doesn't invalidate the token](https://github.com/expressjs/cookie-session/issues/106)
* [Everything you need to know about the `passport-local` Passport JS Strategy](https://levelup.gitconnected.com/everything-you-need-to-know-about-the-passport-local-passport-js-strategy-633bbab6195)


maxlath:
The last added commits propose to add a mechanism to auto-rotate keys

server/lib/auto_rotated_keys.js
```diff
@@ -1,8 +1,19 @@
-const { cookieMaxAge, autoRotateKeys: leadingServer } = require('config')
-const __ = require('config').universalPath
+// This module recovers keys in config/.sessions_keys, and creates and keeps updated
+// an array of keys that will be passed to auth middlewares to sign cookies
+
+// There are presently no automated tests for this behavior, but manual tests can be run
```
jum-s, line 4: maybe some tests could be derived from `tests/api/auth/logout` ?
also the aforementioned test is red on my setup (and green on `master`)

maxlath: * logout test fixed in 7ac297e
* added some tests on sessions in e26dace

jum-s (approved):

PR #459, opened 2020-11-18
replace /api/entities?action=changes by a generalized version of /api/entities?action=contributions

maxlath:
needed by the client for https://github.com/inventaire/inventaire-client/pull/239


server/controllers/entities/contributions.js
```diff
@@ -6,7 +6,7 @@ const responses_ = __.require('lib', 'responses')
 const patches_ = require('./lib/patches')
 
 const sanitization = {
-  user: {},
+  user: { optional: true },
   limit: { default: 100, max: 10000 },
```
jum-s, line 10: maybe 10 000 is too much. It made sense for a single entity, but for all of them, that's only a fraction of the exiting changes and, at the same time, this page could quickly become quite intense for the server to load all those changes.
bonus point : define recent in the branch name `recent-changes` :)

maxlath: @jum-s how much then? 1000?

jum-s: there is a reason why i did not suggest any figure. I guess 1000 is fine

maxlath: done in 52b7368e

Issue #456, opened 2020-11-16
Enable open groups


jum-s:
Like [mobilizon](https://framagit.org/framasoft/mobilizon/-/releases/1.0.1) recently did, group settings would have an option to pick if the group is open to new members or on invite only (like today's implementation).

Default would be on invite.
In case of an open group, there would be no request process, users would become member when pressing "Join the group" (today named "Request to join the group").

maxlath:
Some thoughts:
* I imagine that open groups will strenghten the need for finer visibility settings: having no way to show certain items or shelves to friends but not to members of an open group is getting very sub-optimal. That basically mean that as long as you are part of an open group, I can unilaterally decide to see as much of your inventory as your friends.
* [applies to non-open groups, but might be more critical on open groups] Should we set a maximum number of members? That can wait but I guess we might experience performance issues in both the server and client with very large groups.
* We will need some UI elements to make those settings (`open` but also `searchable`) known to people before they join a group
* Changes on those settings by a group admin should trigger a notification for all members

jum-s:
i agree on the notification part.
I will work on the ui, im thinking of triggering an alert modal when clicking on the "join group"  button. A text that could say "since this group is open, every users can freely join this group and see your items which have a "friends and groups" sharing settings. Would you like to continue and join this group ?" 
On the visibility settings, i think that's par of other issues. Same for the maximum group member (which we could setup when the server is too stressed ?)

maxlath:
done in https://github.com/inventaire/inventaire/pull/460 and https://github.com/inventaire/inventaire-client/pull/240

PR #460, opened 2020-11-18
implement open groups

jum-s:
based on issue #456

### DIFF
--- a/server/controllers/groups/create.js
+++ b/server/controllers/groups/create.js
@@ -13,13 +13,17 @@ const sanitization = {
   searchable: {
     optional: true,
     generic: 'boolean'
+  },
+  open: {
+    optional: true,
+    generic: 'boolean'
   }
 }
 
 module.exports = (req, res) => {
   sanitize(req, res, sanitization)
   .then(params => {
-    const { name, description, position } = params
+    const { name, description, position, open } = params
     let { searchable } = params
 
     if (searchable == null) searchable = true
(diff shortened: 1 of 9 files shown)


server/controllers/groups/create.js
```diff
@@ -29,7 +33,8 @@ module.exports = (req, res) => {
       description: description || '',
       searchable,
       position: position || null,
-      creatorId: req.user._id
+      creatorId: req.user._id,
+      open: open || false
```
maxlath, line 37:
```suggestion
      open
```
There should be no need to re-parse the boolean value, `sanitize` is already doing it. It could be worth making explicit the boolean default default value (`false`), by setting `default: false` in the `sanitization` object (but then do it too for the `searchable` flag)

tests/api/groups/create.test.js
```diff
@@ -71,4 +71,11 @@ describe('groups:create', () => {
       err.body.error_name.should.equal('invalid_name')
     }
   })
+
+  it('should should create an open group', async () => {
```
maxlath, line 75:
```suggestion
  it('should create an open group', async () => {
```

maxlath (changes requested):

server/controllers/groups/create.js
```diff
@@ -13,13 +13,17 @@ const sanitization = {
   searchable: {
     optional: true,
     generic: 'boolean'
+  },
+  open: {
+    optional: true,
+    generic: 'boolean'
```
maxlath, line 19:
```suggestion
    generic: 'boolean'
    default: false
```
my mistake, boolean parameter have no default default value, so it need to be set: the tests fail otherwise

maxlath (changes requested):

maxlath (approved):

PR #462, opened 2020-11-28
Elasticsearch upgrade

jum-s:
upgrade to elasticsearch 7.9
- drop index typing
- integrate elastic-search-engine and couch2elastic4sync lib into inventaire server
- use an explicit mapping (do not rely on `_all` fields match and automatic mappings anymore)
- add an autocomplete analyzer (in `settings`) at indexation to avoid using costly `match_phrase` query at search time
- index only i18n active languages in wikidata entities index (and add a flattened version as string of other languages to still be able to query those)
- index entities popularity in elasticsearch and boost results internally


server/controllers/tasks/lib/search_entity_duplicates_suggestions.js
```diff
@@ -1,22 +1,14 @@
...
-  .catch(_.ErrorRethrow(`${name} search err`))
-}
+  const results = await typeSearch({ search: name, types: [ 'humans' ], filter: 'wd' })
 
-const formatResult = result => ({
-  _score: result._score,
-  uri: prefixifyWd(result.id)
-})
+  return results
+  .filter(result => result._score > 4)
```
jum-s, line 12: maybe we can highlight this arbitrary score filter by making it a constant at the beginning of the file?

server/lib/elasticsearch.js
```diff
@@ -7,27 +7,19 @@ const { host: elasticHost } = CONFIG.elasticsearch
...
-    }
 
     const body = queryBuilder(query, params)
 
-    return requests_.post(customUrl, { body })
+    return requests_.post(url, { body })
     .then(parseResponse)
     .catch(formatError)
-    .catch(_.ErrorRethrow(`${index} ${type} search err`))
+    .catch(_.ErrorRethrow(`${index} search err`))
```
jum-s, line 22: nice catch. Did i forget it because of an absence of tests ?

server/controllers/search/lib/entities_query_builder.js
```diff
@@ -70,7 +70,10 @@ const defaultEntitiesFields = userLang => {
   const fields = [
     'labels.*^2',
     'aliases.*^2',
-    'descriptions.*'
+    'descriptions.*',
+    // Boost titles and subtitles set in claims only
```
jum-s, line 74: why boost of 4 instead of 2 like labels ? wouldnt it favorise to much local collection entities compared to works entities for example?

maxlath: @jum-s that's a bit of a dirty hack to get the `search:entities: should return a local collection` test to fail less often >< The `collectionLabel` is generated using the `randomWords` function, which returns words that regularly happen to be for at least one of them in Wikidata collections labels, which then all get very high score, and push the local collection beyond the 100s rank... There seem to be an issue with the exact match on all words

server/controllers/tasks/lib/search_entity_duplicates_suggestions.js
maxlath: addressed in a83eebaf

server/controllers/search/lib/entities_query_builder.js
maxlath: I removed the commit that added that boost (f215be88), and changed the test instead (35f44049)

jum-s: this is a fine tuning comment but maybe we can depreciate a bit de wikidata boost as you suggested some times ago (but not deleting it yet, there are still to many inv duplicates imho)

maxlath: done (10ce565d), let's see how it goes

maxlath: done (e419c049)

Issue #410, opened 2020-08-05
replace reportPossibleDuplicate email tickets by a task document in database


maxlath:
example: currently, [when a "possible work duplicate" is identified, an email is sent to admins](https://github.com/inventaire/inventaire-client/blob/9d4f8e7/app/modules/entities/views/editor/lib/create_edition_entity_from_work.coffee#L47-L49)

This means that only the admins get the info, and, due to the email setup, an admin wouldn't know if an other admin already had a look at the issue.

This could process could be improved by storing this kind of ticket in the database instead of only sending emails. I guess that could be a type of [task](https://github.com/inventaire/inventaire/blob/4303efa/server/models/task.js), but we would need to decouple tasks from their current unique use-case to a more flexible model, to make it re-usable for a more divers kind of case while keeping what makes a task a task, namely that it needs the attention of an admin or expert user, and that it has resolutions states.

If we were to go that way, maybe those tasks could be scoped to entities data, but that could also be a useful pattern for other non-entities-related administration tasks (ex: user reports)

exilexi:
Yes! I'm not an admin but I am a wikidata enthusiast, I'd be happy to sort out these tickets, and I'm sure other non-admins would too.

maxlath:
related issue: #91

jum-s:
>   what makes a task a task, namely that it needs the attention of an admin or expert user

Today what inventaire consider to be a task is *only* a reconciliation between two authors suspected to represent the same person. A `task` has a `suspectUri` and a `suggestionUri` at its concept. For the issue's usecase, maybe we don't even need to change the task document much, as it is also a reconciliation between 2 entities.

I dont see any reason to includes any kind scoring attributes, which would probably mean some model validations to tweak to accept a task as only a type, a suspect and a suggestion. 

I also dont see any reason to store which user have reported the issue since its an automated report which is sent to admins. If necessary server logs are always around.

jum-s:
[...]

PR #463, opened 2020-11-30
Tasks decaffeinate

jum-s:
improving developer happiness : 
- tests refactor to auto index necessary wikidata entities
- delete collect entities test which was xited for months anyway
- extract create task to be easily used by issue #410


tests/api/utils/entities.js
```diff
@@ -18,6 +22,21 @@ const entitiesUtils = module.exports = {
...
+  findOrIndexEntities: async (uris, index = 'wikidata') => {
+    const ids = _.map(uris, unprefixify)
+    const results = await Promise.all(ids.map(entitiesUtils.getElasticEntity(index)))
+    const entitiesFound = _.filter(results, _.property('found'))
+    const entitiesFoundUris = entitiesFound.map(_.property('_source.uri'))
+    const entitiesNotFoundUris = _.difference(uris, entitiesFoundUris)
+    if (_.isNonEmptyArray(entitiesNotFoundUris)) {
+      // index entities into elasticsearch by getting the uris
+      await entitiesUtils.getByUris(entitiesNotFoundUris)
+      await wait(elasticsearchUpdateDelay)
```
maxlath, line 34:
```suggestion
      await Promise.all(ids.map(id => waitForIndexation('wikidata', id))
```
`waitForIndexation` (introduced in 5667a47a, so accessible after a rebase on `master`) is more reliable than an arbitrary `elasticsearchUpdateDelay`

tests/api/tasks/hooks.test.js
```diff
@@ -9,16 +9,20 @@ const { createTask } = require('../fixtures/tasks')
...
+      // Tests dependency: having a populated ElasticSearch wikidata index
+      const wikidataUris = [
+        'wd:Q535', 'wd:Q54551995', // some Victor Hugos
+        'wd:Q3182477', 'wd:Q228024', // some John Smiths
+        'wd:Q237087' // Fred Vargas
+      ]
       await findOrIndexEntities(wikidataUris)
     })
     it('should update same suspect tasks to merged state', done => {
       // Alexander Kennedy is expected to have several merge suggestions
```
maxlath:
```suggestion
      // Victor Hugo is expected to have several merge suggestions
```

server/controllers/tasks/lib/get_new_tasks.js
```diff
@@ -15,21 +17,87 @@ module.exports = entity => existingTasks => {
...
-const filterOutExistingTasks = existingTasks => suggestions => {
+const filterNewTasks = existingTasks => suggestions => {
   const existingTasksUris = _.map(existingTasks, 'suggestionUri')
   return suggestions.filter(suggestion => !existingTasksUris.includes(suggestion.uri))
 }
+
+const authorNameInWorkTitles = (authorTerms, workLabels) => {
+  for (const authorLabel of authorTerms) {
+    for (const workLabel of workLabels) {
+      return workLabel.match(authorLabel)
```
maxlath, line 44: this makes the loop return the match of the first `authorLabel` with the first `workLabel` and nothing else

```suggestion
      if (workLabel.match(authorLabel) != null) then return true
```

server/controllers/tasks/lib/get_new_tasks.js
```diff
@@ -16,16 +15,15 @@ module.exports = (entity, existingTasks) => {
...
-    .then(buildTasksObjects(suspectUri, 'deduplicate'))
   })
 }
 
-const evaluateSuggestions = (suspect, workLabels) => suggestions => {
+const filterOrMergeSuggestions = (suspect, workLabels) => suggestions => {
   const suspectTerms = getEntityNormalizedTerms(suspect)
   // Do not automerge if author name is in work title
-  // as it confuses occurences finding on WP pages
+  // as it confuses occurences found wikipedia pages
```
maxlath, line 28:
```suggestion
  // as it confuses occurences found on Wikipedia pages
```

maxlath (changes requested):

server/controllers/tasks/lib/get_new_tasks.js
jum-s: [documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/match) says:

> `.match` returns an Array or null if no matches are found.

since arrays are truthy, this is what we need here. I mean that even when .match does not return a boolean, the if where this function is consumed understands it this way.

this PR does not modify this function, [its the same as before](https://github.com/inventaire/inventaire/blob/master/server/controllers/tasks/lib/evaluate_suggestions.js#L19)

maxlath: it's not the same as before, there was an `if` before (I shouldn't have added the `!= null` part in the suggestion above, as it's indeed irrelevant)

jum-s: [forcing myself to not continue this conversation :) ]

PR #468, opened 2020-12-19
decouple patches ids from entities revs

maxlath:
After restoring the entities database without documents revs ids, all entities got their `_rev` integer component reset to `1`, which meant that new edits on those entities triggered the creation of patches with the same id as an existing patch, which is thus rejected as a conflict.

This PR addresses the problem by creating a version number on entities independent from revs ids.

After merging, and **before** deploying:
- set a `version` number on all entities
- try to recover lost patches


jum-s (approved):
approving but considering 2 more suspicious files:


server/controllers/entities/lib/revert_merge.js#17
scripts/dumps/lib/serialize_entity_in_turtle.js#23

PR #469, opened 2020-12-22
update geo queries

maxlath:
* replace LevelDB geospatial search by Elasticsearch [`geo_distance`](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-geo-bounding-box-query.html) queries: removing the need to maintain a geo index of users in LevelDB
* replace CouchDB `byGeoSquare` views queries by Elacticsearch [`geo_bounding_box`](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-geo-bounding-box-query.html) queries: way less hacky and allows more precision in the results

The primary motivation of this PR was to get rid of the geo index, as, following the server migration, secondary indexes are expected to be initialized by scripts and then only kept in sync by the server `follow` lib: by getting rid of the geo index, we get rid of the need to write a script to reinitialize from an existing primary db; the work is already done when reindexing documents in Elasticsearch \o/


tests/unit/libs/046-sanitize.js
```diff
@@ -460,4 +460,43 @@ describe('sanitize', () => {
...
+      const req = { query: { bbox: '[0, 2, 1, 1]' } }
+      const configs = { bbox: {} }
+      await sanitize(req, {}, configs)
+      .then(shouldNotBeCalled)
+      .catch(err => {
+        err.message.should.startWith('invalid bbox')
+      })
+    })
+
+    it('should reject bbox with a higher minLat than maxLat', async () => {
```
jum-s, line 485: so no limit. One can query the whole world theoretically?

tests/api/fixtures/users.js
```diff
@@ -88,6 +89,15 @@ const API = module.exports = {
   getTwoFriends: () => {
     twoFriendsPromise = twoFriendsPromise || getTwoFriends()
     return twoFriendsPromise
+  },
+
+  getRandomPosition: () => {
```
jum-s, line 94: maybe some DRYing with https://github.com/inventaire/inventaire/blob/master/tests/api/utils/users.js#L6

server/lib/search_by_position.js
```diff
@@ -0,0 +1,43 @@
...
+// See https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-geo-bounding-box-query.html
+const queryBuilder = ([ minLng, minLat, maxLng, maxLat ]) => {
+  const topLeft = { lat: maxLat, lon: minLng }
+  const bottomRight = { lat: minLat, lon: maxLng }
+  return {
+    query: {
+      bool: {
+        must: {
+          match_all: {}
+        },
```
jum-s, line 30: is an empty `match_all` necessary ?

tests/unit/libs/046-sanitize.js
maxlath: there is a hard coded limit of 500 in `server/lib/search_by_position.js`, so searching the whole world is possible, but only the first 500 results will be returned. We could pass it a `limit` parameter from the `search_by_position` endpoint

tests/api/fixtures/users.js
maxlath: done in geo-queries

server/lib/search_by_position.js
maxlath: that's what the documentation proposes https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-geo-bounding-box-query.html
but it works when removing it (2bb13766 and e2215d8d), as long as we don't set a `minScore`, as the `filter` doesn't give any matching score (`hit._score = 0`)

jum-s (approved):

PR #470, opened 2020-12-24
get existing entities by isbn

jum-s:
pass a boolean param (`existing`) to get_entity_by_uris to not trigger any creation when the requester does not want to create any entity when passing an isbn
known cases: checking if entity exists and/or if it has a cover before
uploading the image in order to feed the resolver with an `"invp:P2" : "imageHash"` from the media container only when necessary

### DIFF
--- a/server/controllers/entities/by_uris_get.js
+++ b/server/controllers/entities/by_uris_get.js
@@ -14,6 +14,10 @@ const validRelativesProperties = [
 const sanitization = {
   uris: {},
   refresh: { optional: true },
+  autocreate: {
+    generic: 'boolean',
+    optional: true
+  },
   relatives: {
     allowlist: validRelativesProperties,
     optional: true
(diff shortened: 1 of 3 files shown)


maxlath:
* That should maybe be the default behavior, and the autocreation from dataseed could then require a flag (`dataseed=true` or `autocreate=true`?)
* Reusing the `missingIsbns.length === 0` early return prevents the creation of a `notFound` array: there could be a test that requesting an entity by an unknown isbn without dataseed (`existing=true` or `autocreate=false`) should return the request uri in the `notFound` array

jum-s:
related issue #481

jum-s:
About naming: "creating" or "existing" is clearer to understand than any dataseed reference imo. Maybe we could use the `dry` flag, already wildly used in the application.
About default behavior: the original intend was to not change the api behavior (thinking about people relying on the api today). I understand its not breaking change, yet it is a change. Maybe something like "if there is a dataseed service available use it, otherwise skip" behavior would be better?
About tests: good point, I will add some tests which are non existent on this part of the app now

jum-s:
finally adopted the autocreate true as default to move toward less dataseed slowly.
see also client branch https://github.com/inventaire/inventaire-client/pull/257

maxlath (approved):

PR #480, opened 2021-01-04
Send notification on group settings update

maxlath:
Addresses the point from https://github.com/inventaire/inventaire/issues/456#issuecomment-729921968 on notifications

* send notification on group settings update
* update existing unread notifications instead of spamming with notifications if a setting is changed several times (update existing notification), or changed an reverted (delete existing notification)
* do some cleanup around notifications:
  * reorganize as a controller folder and split lib file
  * run operations in bulk 

~NB: this PR makes use of a feature yet unused from the recently upgraded CouchDB: `_find`. See 8f4b53f commit message. This could totally have been done with a classic view, but it was a good occasion to test this new capability.~ [the mango query was replaced by a classic view query for the sake of simplicity. Keeping the exploration in the git history to make it usable in future, more justified, uses]

To do after merging this PR:
- add support for those new notifications on the client


tests/api/groups/update_settings.test.js
```diff
@@ -97,4 +90,108 @@ describe('groups:update-settings', () => {
...
+        value: true
+      })
+      const notifications = await getNotifications({ user: member, type: 'groupUpdate', subject: groupId })
+      const notification = notifications.find(isAttributeNotification('searchable'))
+      notification.data.attribute.should.equal('searchable')
+      notification.data.previousValue.should.equal(false)
+      notification.data.newValue.should.equal(true)
+    })
+
+    it('should send a notification to groups members when updating the open flag', async () => {
```
jum-s, line 142: i dont think testing every single attributes is necessary (even though it was nice for development purposes), now its feels its only redundant tests that takes space. Maybe it could be unit tests instead
On this file maybe it would be nice to test what happen if attribute's value if of a wrong type, ie. what kind of error shall it throw

server/models/validations/notification.js
```diff
@@ -0,0 +1,32 @@
...
+const dataValidationPerType = {
+  friendAcceptedRequest: ({ user }) => {
+    validations.pass('userId', user)
+  },
+  groupUpdate: ({ group, user, attribute, previousValue, newValue }) => {
+    validations.pass('groupId', group)
+    validations.pass('userId', user)
+    passFromFunction('attribute', attribute, _.isString)
+    Group.validations.pass(attribute, previousValue)
+    Group.validations.pass(attribute, newValue)
```
jum-s, line 26: those have already been validated by the group itself, when it was updated no? I dont think its the concern of the Notification model to know how is group updated (otherwise notification will end up knowing more models)

server/models/validations/notification.js
```diff
@@ -0,0 +1,32 @@
...
+}
+
+const dataValidationPerType = {
+  friendAcceptedRequest: ({ user }) => {
+    validations.pass('userId', user)
+  },
+  groupUpdate: ({ group, user, attribute, previousValue, newValue }) => {
+    validations.pass('groupId', group)
+    validations.pass('userId', user)
+    passFromFunction('attribute', attribute, _.isString)
```
jum-s, line 24: that's new. is there a reason to not have a simple `  string: _.isString,` in `validations/common.js` ?

server/models/validations/notification.js
maxlath: here the Notification model doesn't know anything else than this attribute is known by the Group model, which should be able to confirm that it looks right before we sent it to the database

But I'm open to suggestions to replace it, including deleting it

server/models/validations/notification.js
maxlath: I'm not exactly sure now, but I think the main motive was to experiment with more direct way to validate, that would not rely on this `this` witchcraft, giving more flexibility if the time comes were an other notification type uses the `attribute` field differently... but yeah `attribute` will always be a string so maybe that's not the best example

jum-s: true, it was a syntax question but youre right `attribute` is completely controlled by the system, there is no exterior input. I would delete this line.

server/models/validations/notification.js
jum-s: So in the end, every value in a model will end up being re-checked when a new notification is created.
What happens if this notification validation dont pass ? As far as i understand, the action which triggered the notification would be complete (a group action here) but the notification wont be created.
I think Notification model have to trust who is triggering it. What it must know though is that if value is a string in order to properly display it in the client.

I would delete those two lines.

server/models/validations/notification.js
jum-s: also, as a purely opinionated comment : meta programming such as this confuses newbies by making is harder to read (and i say that coming from the ruby world...)

server/models/validations/notification.js
maxlath: imo that's the role of the model to assert that what is put in the database conforms to certain expectations (expectations which are much needed to not make NoSQL databases more chaotic than they already are), and to throw errors at consumers that don't respect those expectations, allowing maintainers to correct those consumers' behavior: ideally those errors would be detected and fixed during development, but if happening in prod, it's preferable to throw an error at the closest point from the problem rather than letting through inconsistent data that would trigger errors down the road when expectations on which db queries build are not met

Also, those tests are very cheap

server/models/validations/notification.js
maxlath: I meant `attribute` in any model will be of type string, so it can be added to `validations/common.js`, but I'm in favor of keeping this line as it guards against undefined/non-string values passing through (same thinking as expressed in https://github.com/inventaire/inventaire/pull/480#discussion_r564382134)

maxlath: `passFromFunction` was removed in 6968834

server/models/validations/notification.js
maxlath: > those have already been validated by the group itself, when it was updated no?

the notification model doesn't know that

jum-s: > it's preferable to throw an error at the closest point from the problem
Totally agree that's why the Group model has its own validation, and validated the data already
A model using another model's validation is opening a lot of unnecessary validations, i disagree on the cheapness, its mixing models, could become spaghettis code which is already the case in the controllers (circular dependency stuff) but model were pretty clean so far. Would be nice to keep it this way.

> the notification model doesn't know that
The notification model doesnt have to know anything else than how to produce a notification. Group validation is not the job of the notification model

maxlath: > The notification model doesnt have to know anything else than how to produce a notification. Group validation is not the job of the notification model

but this validation **is** intended to produce a notification, namely a group update notification, which is a sub-type of notification associated with the group model, just like the patch model is associated with the entity model, or the transaction model with the item model; calling elements of validation from that associated model makes explicit that association and doesn't chock me, au contraire

> A model using another model's validation is opening a lot of unnecessary validations
runtime assertions (just like all those `assert_.string`) are always unnecessary, rechecking the obvious, the already validated upstream, until they are not because they help you detect a bug before it makes it to the database: I like runtime assertions

I would thus still prefer to keep those, but wouldn't make this issue a blocking point, so if this is blocking for you, I can remove it (with a hint of grunting, but that would be quickly forgotten)

jum-s: after thinking about it, i think its necessary considering functional programming cannot ensure data consistency. In those cases, i feel that OOP is so much superior to keep validations at its place only.
I still dont feel good about mixing types of validations, but i guess i just have to swallow it and grumbling about FP

jum-s (approved):

Issue #313, opened 2019-09-30
locally keep track of Wikidata entity relations created in Inventaire


maxlath:
so that we don't have to wait on the Wikidata Query Service update to display the relation

Use-cases:
* you move a Inventaire work with a Wikidata author to Wikidata, then go to the author's page:
* you add a serie to a Wikidata work entity, then go on the serie's page

In both cases, the work doesn't appear because the query that request the serie's parts find outdated data on Wikidata Query Service (the data will be updated somewhere in the following hour)

Problem: yet another cache to invalidate? which source of truth to trust and when?

maxlath:
we could have a service that takes care of:
- adding the known missing triples to a cache
- refresh the known invalid queries after a delay:
  - if it finds the cached triple, cleanup local cache
  - if it doesn't find it, reschedule this check
- give up after a certain number of attempts or excessive time, and cleanup local cache: that could address cases where the data was later modified so that our locally cached triple has been removed from Wikidata and it's absence in WQS results is thus not a problem of data freshness anymore

maxlath:
solved by #417

PR #488, opened 2021-01-20
more sophisticated invalidation of Wikidata Query Service cached results

maxlath:
In the current situation:
- a user sets a P50 on a Wikidata entity from the Inventaire entity editor
- the [entities relations temporary cache](https://github.com/inventaire/inventaire/blob/a85f27d/server/controllers/entities/lib/temporarily_cache_relations.js) keeps that relation in memory for a certain time, so that when some user requests the entities having a P50 claim pointing to that same target entity, we can add to the result the value we are aware of, before the Query Service update
- after that time (4h by default), the relation is removed from the temporary cache **but** cached queries pre-dating the initial update might still be in the cache (can be seen with `lev ./db/level-tests --prefix '!cache!wdQuery'`), returning results we know to be out-dated

This PR makes sure that those cached results are invalidated once the temporary cache expired, that is, after we can reasonably expect that the Query Service is now up-to-date

[centralizing issue on the topic: #313 ]

### DIFF
--- a/server/data/wikidata/run_query.js
+++ b/server/data/wikidata/run_query.js
@@ -1,10 +1,12 @@
 const __ = require('config').universalPath
 const _ = __.require('builders', 'utils')
+const radio = __.require('lib', 'radio')
 const cache_ = __.require('lib', 'cache')
 const error_ = __.require('lib', 'error/error')
 const wdk = require('wikidata-sdk')
 const makeSparqlRequest = require('./make_sparql_request')
-const queries = require('./queries/queries')
+const { queries, queriesPerProperty } = require('./queries/queries')
+const { unprefixify } = __.require('controllers', 'entities/lib/prefix')
 const possibleQueries = Object.keys(queries)
 const dashesPattern = /-/g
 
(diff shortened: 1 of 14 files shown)


server/data/wikidata/run_query.js
```diff
@@ -54,9 +62,18 @@ const parametersTests = {
...
-  const sparql = queryBuilder(params)
+  const sparql = queries[queryName].query(params)
 
   return makeSparqlRequest(sparql)
   .catch(_.ErrorRethrow(key))
 }
+
+radio.on('invalidate:wikidata:entities:relations', async ({ property, valueUri }) => {
+  const queriesToInvalidate = queriesPerProperty[property] || []
+  queriesToInvalidate.push(...queriesPerProperty['*'])
```
jum-s, line 73: this `'*'` looks cryptic, could we leave a comment explaining what does this do ?

jum-s (approved):

server/data/wikidata/run_query.js
maxlath: Indeed. This is used for queries that should be invalidated for any value of 'property': would writing that in a comment be enough or is there also a better way to express that than `*`?

PR #490, opened 2021-01-23
move_to_wikidata: add context to the redirection patch

maxlath:
to not be surprised each time a non-dataadmin user makes a merge this way

To fully complete its purpose, we would also need to have the client display patches context in history views

### DIFF
--- a/server/controllers/entities/lib/move_to_wikidata.js
+++ b/server/controllers/entities/lib/move_to_wikidata.js
@@ -21,7 +21,14 @@ module.exports = async (user, invEntityUri) => {
   // by the absence of the entity they just moved to Wikidata in lists generated with the help of the WQS
   await cacheEntityRelations(invEntityUri)
 
-  await mergeEntities({ userId: reqUserId, fromUri: invEntityUri, toUri: wdEntityUri })
+  await mergeEntities({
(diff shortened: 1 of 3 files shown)


server/controllers/entities/lib/move_to_wikidata.js
```diff
@@ -21,7 +21,15 @@ module.exports = async (user, invEntityUri) => {
...
   await cacheEntityRelations(invEntityUri)
 
-  await mergeEntities({ userId: reqUserId, fromUri: invEntityUri, toUri: wdEntityUri })
+  await mergeEntities({
+    userId: reqUserId,
+    fromUri: invEntityUri,
+    toUri: wdEntityUri,
+    context: {
+      action: 'move-to-wikidata',
+      entityId
```
jum-s, line 30: invEntityUri === inv:${entityId}

so patches store twice the "from entity id" no ?

maxlath: yes, that could be removed

maxlath: fixed in 508e9522

jum-s (approved):

PR #498, opened 2021-02-16
find entities with terms longer than autocomplete max ngram

### LINKED ISSUE #485: Local entities are easily not found despite an exact match
when their label share some words with Wikidata entities (despite the absence of popularity on those)

maxlath:
fix #485 last issue, or at least makes the last failing test pass (without adding any new field or requiring a reindexation \o/)

We could still imagine cases where the exact match would have false positives, when terms are longer than 10 characters. For instance searching "Naturwissenschaft" would consider the following as exact matches:
- "Naturwissenschaftler"
- "Naturwissenschaftlerin"
- "naturwissenschaftlich"
- or just any term that matches `/naturwisse\w*/i`

But that looks like an acceptable compromise, as most `exact` requests involve several words (ex: "Metaphysische Anfangsgründe der Naturwissenschaft") that all have to be found in the field

TODO after merge:
- [update prod Elasticsearch indexes with the new settings](https://github.com/inventaire/inventaire/blob/4ebe2ae/server/db/elasticsearch/settings/settings.js#L3-L10)

### DIFF
--- a/server/db/elasticsearch/settings/settings.js
+++ b/server/db/elasticsearch/settings/settings.js
@@ -1,5 +1,14 @@
 // See: https://www.elastic.co/guide/en/elasticsearch/reference/7.10/search-analyzer.html
 
+// To update the settings of an existing index:
+// - Close index
+//   curl -XPOST ${elastic_host}/${index_name}/_close
+// - Update index settings:
+//   settings_json=$(node -p 'JSON.stringify(require("./server/db/elasticsearch/settings/settings.js"))')
+//   curl -XPUT ${elastic_host}/${index_name}/_settings -d "$settings_json"
+// - Reopen index
+//   curl -XPOST ${elastic_host}/${index_name}/_open
+
 const maxGram = 10
 
 module.exports = {
--- a/server/db/elasticsearch/settings/settings.js
+++ b/server/db/elasticsearch/settings/settings.js
@@ -21,6 +21,10 @@ module.exports = {
         // See https://www.elastic.co/guide/en/elasticsearch/reference/7.10/analysis-standard-tokenizer.html
         tokenizer: 'standard',
         filter: [
+          // The 'standard' tokenizer does not imply lowercase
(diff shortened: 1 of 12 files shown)


server/db/elasticsearch/settings/settings.js
```diff
@@ -21,10 +38,24 @@ module.exports = {
         // See https://www.elastic.co/guide/en/elasticsearch/reference/7.10/analysis-standard-tokenizer.html
         tokenizer: 'standard',
         filter: [
+          // The 'standard' tokenizer does not imply lowercase
+          // https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-standard-tokenizer.html
+          // not to be confused with the 'standard' analyzer, which does
```
jum-s, line 43: ahah tricky indeed

tests/api/search/entities.test.js
```diff
@@ -84,18 +84,20 @@ describe('search:entities', () => {
...
+        const label = "L'EAU DOUCE EN PÉRIL"
+        const work = await createWork({ labels: { fr: label.toLowerCase() } })
+        await waitForIndexation('entities', work._id)
+        const results = await search({ types: 'works', search: label, lang: 'fr', exact: true })
+        _.map(results, 'uri').should.containEql(work.uri)
+      })
+
+      it('should find a label containing terms longer than the autocomplete max ngram', async () => {
+        // Current max_ngram=10
         const label = 'Metaphysische Anfangsgründe der Naturwissenschaft'
```
jum-s, line 97: since `max_ngram` is now a constant in `server/db/elasticsearch/settings/settings.js`, what about requiring it inside this test and generate a label with words longer than this limit instead.
```suggestion
        const long_word = () => randomWords(10).split(' ').join('').slice(0, max_ngram + 1)
        const label = `${long_word()} ${long_word()} ${long_word()}` 
```

maxlath: done in 869d0fe5

jum-s (approved):

PR #499, opened 2021-02-16
fix popularity boost in entities search

maxlath:
Due to the `boost_mode=sum`, entities that have a high popularity some how pop up in results despite there lexical match being very low. Example: when searching "Jean Baudrillard", wd:Q182847 should arrive first, and certainly not arrive after the (glorious but irrelevant) René Goscinny (wd:Q192214) ahead of him

This PR proposes to come back to `modifier=ln2p` and the default `boost_mode=multiply`


jum-s (approved):

jum-s:
note that `jean baudrillar` (without ending d) have [very satisfying results](https://inventaire.io/api/search?types=humans&search=jean%20baudrillar&limit=10) despite a low popularity

Issue #506, opened 2021-02-20
resolver does not update date when more precise


jum-s:
when i use resolve some entries with the update flag, and the entries have more precise date (`2010-01-01`) than the current entity (`2010`) then it doesnt update the entity date.

jum-s:
proposed solution: add an entity update around [update_resolved_entry](https://github.com/inventaire/inventaire/blob/master//server/controllers/entities/lib/resolver/update_resolved_entry.js#L26) for 'simple-day' [properties](https://github.com/inventaire/inventaire/blob/master//server/controllers/entities/lib/properties/properties_values_constraints.js) which would check some simple heuristic : the larger number of digits in a `simple-day` format wins

PR #507, opened 2021-02-20
Resolver update date

jum-s:
solve #506 
if entry date is more precise, update the entity

### DIFF
--- a/server/controllers/entities/lib/resolver/update_resolved_entry.js
+++ b/server/controllers/entities/lib/resolver/update_resolved_entry.js
@@ -22,8 +22,7 @@ const updateEntityFromSeed = (reqUserId, batchId) => async seed => {
   if (prefix === 'wd') return
 
   const entity = await getEntity(prefix, entityId)
-  await updateDatePrecision(entity, seedClaims, reqUserId, batchId)
-  await addMissingClaims(entity, seedClaims, imageUrl, reqUserId, batchId)
+  await updateClaims(entity, seedClaims, imageUrl, reqUserId, batchId)
 }
 
 const getEntity = (prefix, entityId) => {
@@ -34,13 +33,18 @@ const getEntity = (prefix, entityId) => {
   }
 }
 
-const addMissingClaims = async (entity, seedClaims, imageUrl, reqUserId, batchId) => {
-  // Do not update if property already exists
+const updateClaims = async (entity, seedClaims, imageUrl, reqUserId, batchId) => {
+  // Do not update if property already exists (except if date is more precise)
   // Known cases: avoid updating authors who are actually edition translators
+  const updatedEntity = _.cloneDeep(entity)
   const newClaims = _.omit(seedClaims, Object.keys(entity.claims))
+  _.forEach(newClaims, (newValues, prop) => {
+    updatedEntity.claims[prop] = newClaims[prop]
+  })
+  updateDatePrecision(entity, updatedEntity, seedClaims)
   await addImageClaim(entity, imageUrl, newClaims)
-  if (_.isEmpty(newClaims)) return
-  return entities_.addClaims(reqUserId, newClaims, entity, batchId)
+  if (_.isEqual(updatedEntity, entity)) return
+  await entities_.putUpdate({ userId: reqUserId, currentDoc: entity, updatedDoc: updatedEntity, batchId })
 }
 
 const addImageClaim = async (entity, imageUrl, newClaims) => {
@@ -51,9 +55,8 @@ const addImageClaim = async (entity, imageUrl, newClaims) => {
   newClaims['invp:P2'] = [ imageHash ]
 }
 
-const updateDatePrecision = async (currentDoc, seedClaims, userId, batchId) => {
+const updateDatePrecision = (currentDoc, updatedDoc, seedClaims) => {
   const seedDateClaims = _.pick(seedClaims, simpleDayProperties)
-  const updatedDoc = _.cloneDeep(currentDoc)
 
   _.forEach(seedDateClaims, (seedDates, prop) => {
     const seedDate = seedDates[0]
@@ -62,8 +65,6 @@ const updateDatePrecision = async (currentDoc, seedClaims, userId, batchId) => {
       updatedDoc.claims[prop] = seedDateClaims[prop]
     }
   })
-  if (_.isEqual(updatedDoc, currentDoc)) return
-  await entities_.putUpdate({ userId, currentDoc, updatedDoc, batchId })
 }
 
 const simpleDayProperties = [ 'wdt:P569', 'wdt:P570', 'wdt:P571', 'wdt:P576', 'wdt:P577' ]
--- a/server/controllers/entities/lib/resolver/update_resolved_entry.js
+++ b/server/controllers/entities/lib/resolver/update_resolved_entry.js
@@ -38,7 +38,7 @@ const updateClaims = async (entity, seedClaims, imageUrl, reqUserId, batchId) =>
   // Known cases: avoid updating authors who are actually edition translators
   const updatedEntity = _.cloneDeep(entity)
   const newClaims = _.omit(seedClaims, Object.keys(entity.claims))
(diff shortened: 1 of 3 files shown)


server/controllers/entities/lib/resolver/update_resolved_entry.js
```diff
@@ -35,13 +35,23 @@ const getEntity = (prefix, entityId) => {
...
 }
 
-const addMissingClaims = async (entity, seedClaims, imageUrl, reqUserId, batchId) => {
-  // Do not update if property already exists
+const updateClaims = async (entity, seedClaims, imageUrl, reqUserId, batchId) => {
+  // Do not update if property already exists (except if date is more precise)
   // Known cases: avoid updating authors who are actually edition translators
+  const updatedEntity = _.cloneDeep(entity)
   const newClaims = _.omit(seedClaims, Object.keys(entity.claims))
+  _.forEach(newClaims, (newValues, prop) => {
```
maxlath, line 43: [minor point: syntax]
I would be of the opinion that we should stick to standard JS options when available, rather than libs
```suggestion
  Object.keys(newClaims).forEach(prop => {
```
or
```suggestion
  for (const prop in newClaims) {
```

server/controllers/entities/lib/resolver/update_resolved_entry.js
```diff
@@ -51,3 +61,21 @@ const addImageClaim = async (entity, imageUrl, newClaims) => {
   const { url: imageHash } = await getImageByUrl(imageUrl)
   newClaims['invp:P2'] = [ imageHash ]
 }
+
+const updateDatePrecision = (entity, updatedEntity, seedClaims) => {
+  const seedDateClaims = _.pick(seedClaims, simpleDayProperties)
+
+  _.forEach(seedDateClaims, (seedDates, prop) => {
```
maxlath, line 68: [minor point: syntax]
same as above, even if here the 2 arguments bring some more value, the standard way isn't thaaaat much of a stretch:
```suggestion
  Object.keys(seedDateClaims).forEach(prop => {
    const seedDates = seedDateClaims[prop]
```
or
```suggestion
  for (const [ prop, seedDateClaims ] of Object.entries(seedDateClaims) {
```
(that last one takes a bit of getting use to, but it's standard, and learning it means you can go use that knowledge everywhere JS is accepted with zero dependency, that is, to our great joy, everywhere ;))

server/controllers/entities/lib/resolver/update_resolved_entry.js
```diff
@@ -51,3 +61,21 @@ const addImageClaim = async (entity, imageUrl, newClaims) => {
...
+      updatedEntity.claims[prop] = seedDateClaims[prop]
+    }
+  })
+}
+
+const simpleDayProperties = [ 'wdt:P569', 'wdt:P570', 'wdt:P571', 'wdt:P576', 'wdt:P577' ]
+
+const isMorePreciseDate = (date1, date2) => intCount(date1) > intCount(date2)
+
+const intCount = str => (str.match(/\d/g) || []).length
```
maxlath, line 81:
```suggestion
const isMorePreciseDate = (date1, date2) => datePartsCount(date1) > datePartsCount(date2)

const datePartsCount = simpleDay => simpleDay.split('-').length
```

tests/api/entities/resolver/resolve_and_update.test.js
```diff
@@ -182,6 +182,69 @@ describe('entities:resolver:update-resolved', () => {
...
+      }
+    }
+    const { entities: preResolvedEntities } = await getByUris(uri)
+    const preResolvedEntityVersion = Object.values(preResolvedEntities)[0].version
+
+    await resolveAndUpdate(entry)
+    await wait(10)
+    const { entities: postResolvedEntities } = await getByUris(uri)
+    const postResolvedEntityVersion = Object.values(postResolvedEntities)[0].version
+    postResolvedEntityVersion.should.equal(preResolvedEntityVersion + 1)
```
maxlath, line 222: would be nice to also assert that the resulting date is the one expected:
```suggestion
    const postResolvedEntity = postResolvedEntities[uri]
    postResolvedEntity.version.should.equal(preResolvedEntityVersion + 1)
    postResolvedEntity.claims['wdt:P577'].should.deepEqual([ entryDate ])
```

maxlath (changes requested):
looks good. I would like to consider a special case though: if `entityDate = '2020'` and `entryDate = '2021-01'`, the fact that the entry doesn't bring a precision to the existing date, but a disagreement should make the update function fallback to ignoring the more precise date, and stay with the previously set date, no?

server/controllers/entities/lib/resolver/update_resolved_entry.js
jum-s: I dont get it, to me, `2021-01` is more precise than `2020`, why would i be "better" to ignore this update ?

tests/api/entities/resolver/resolve_and_update.test.js
jum-s: this suggestion is, to me, outside the scope of this test, which currently keeps it to "is there an update" aka is there a new version. The more specific the test, the less maintainable (ex: if the date format changes in the future, we would only need to change the hardcoded values at the beggining of this test). imo, this suggestion would fit for a unit test.

server/controllers/entities/lib/resolver/update_resolved_entry.js
jum-s: this would add some significant changes in the code. What about seeing it as: if its more precise, most chances is that its more relevant (since its more difficult to make this up) ?

server/controllers/entities/lib/resolver/update_resolved_entry.js
jum-s: `Object.keys(seedDateClaims).forEach` feels more friendly to people coming from other languages though.

server/controllers/entities/lib/resolver/update_resolved_entry.js
```diff
@@ -49,3 +59,39 @@ const addImageClaim = async (entity, imageUrl, newClaims) => {
...
+  else if (date1Parts.length > date2Parts.length) {
+    mostPreciseDate = date1Parts
+    leastPreciseDate = date2Parts
+  } else {
+    mostPreciseDate = date2Parts
+    leastPreciseDate = date1Parts
+  }
+  const truncatedPreciseDate = mostPreciseDate.slice(0, leastPreciseDate.length)
+  return areDatesPartsEqual(truncatedPreciseDate, leastPreciseDate)
+}
```
maxlath, line 91: the simple-day format, in which month and days are 0-padded, allows to implement `doDatesAgree` as a simple string comparison:
```suggestion
const doDatesAgree = (seedDate, currentDate) => seedDate.startsWith(currentDate)
```

tests/api/entities/resolver/resolve_and_update.test.js
```diff
@@ -181,6 +181,89 @@ describe('entities:resolver:update-resolved', () => {
...
+        claims: { 'wdt:P577': entryDate }
+      }
+    }
+    const { entities: preResolvedEntities } = await getByUris(uri)
+    const preResolvedEntityVersion = Object.values(preResolvedEntities)[0].version
+
+    await resolveAndUpdate(entry)
+    await wait(10)
+    const { entities: postResolvedEntities } = await getByUris(uri)
+    const postResolvedEntityVersion = Object.values(postResolvedEntities)[0].version
```
maxlath, line 220: [detail] we have a `getByUri` function to make this classic pattern less verbose
```suggestion
    const { version: postResolvedEntityVersion } = await getByUri(uri)
```

maxlath (changes requested):

server/controllers/entities/lib/resolver/update_resolved_entry.js
jum-s: ahah indeed ! i wanted to have an more agnostic comparison to compare whatever length on both side, but i guess YAGNI

maxlath (approved):

Issue #187, opened 2018-05-24
implement ActivityPub to enable federated Inventaire instances


maxlath:
**This issue isn't a todo, rather a way to investigate and document a possible future:**

#### Steps

1. Move Inventaire entities to a Wikibase instance (#186), that would be the main non-decentralized service among federated Inventaire instances, the idea being that decentralizing this knowledge base would be pure waste of contributors efforts, that would have to re-do the work on every instances.

2. Turn Inventaire into an ActivityPub-powered social network
  * [ ] turn additions to inventories (currently known as items #180), collections (#184), reviews (#185), transactions actions and messages ActivityPub events
  * [ ] find a solution to the problem of indexing inventories per entity URIs over the fediverse

Other services that could benefit from not being decentralized, or at least to offer the option to use a common instance between several instances:
 * The [entities search engine](https://github.com/inventaire/entities-search-engine): the entities remaining centralized, all running a search engine could just be a waste of resources(?)

#### Pros
* centralizing everything has never been the goal of this project: the vision is still to aggregate information on resources from different providers, be them peers, libraries, or bookshops
* that would probably bring more code contributors, just interested by improving/customizing their instance, giving the opportunity to explore adding new kinds of resources beyond books in a full [bazar](https://en.wikipedia.org/wiki/The_Cathedral_and_the_Bazaar) way
* [decentralization is cool ;)](https://framablog.org/2018/05/23/comment-reparer-les-medias-sociaux-et-faire-encore-mieux/)

#### Cons
* Federation is still difficult to understand for many people used to centralized services
* And then even for tech-saavy people, federation is hard: which instance should I choose, etc
* Quality of service is hard to keep consistent over many instances, run by a diversely qualify crowd of sys admins and/or amateurs. Several issue on Mastodon seem to be linked to this: mismatch between following/followers count and the actual list, etc
* Yeaah let's do all the work, and let a startup deploy it's customized version with ads and affiliation to Amazon on every pages, hurraay...

#### Questions
* Is that even possible? Inventaire isn't "just" a microblogging platform, there are some heavy optimization to do to be able to [display inventory content in a meaningful way](https://github.com/inventaire/inventaire-client/issues/119), etc, how would that work over the fediverse? Discussing this possibility with @Chocobozzz, ActivityPub seems fit to activity streams, not to browsing large inventories over a myriad of instances(?)

#### Requirements
* [ ] full import/export features: it is currently possible to export an inventory, but not to re-import it in a clean/complete way

[...]

PR #508, opened 2021-02-21
Add webfingers

jum-s:
Translate `adamsberg@inventaire.io` mentions to user profile url `https://inventaire.io/user/adamsberg`.
Thanks to a `.well-known/webfinger?resource=acct:adamsberg@inventaire.io` request.
related to #187

### DIFF
--- a/server/controllers/activitypub/webfinger.js
+++ b/server/controllers/activitypub/webfinger.js
@@ -1,0 +1,50 @@
+const CONFIG = require('config')
+const __ = require('config').universalPath
+const error_ = __.require('lib', 'error/error')
+const assert_ = __.require('utils', 'assert_types')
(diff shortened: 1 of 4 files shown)


jum-s:
at some point [this](https://github.com/inventaire/inventaire-deploy/blob/master/nginx/inventaire.original.nginx#L230) may be a problem

jum-s:
WIP since it does seem to work... As far as i understood webfinger are enough for another instance to recognize a user account and display the account results

jum-s:
works now : in order to be recognized an actor needs a valid inbox and a valid outbox (aka answering a GET 200 on inbox and outbox urls). 
Webfinger was not the problem

jum-s:
close as it will be included in the activity pub branch

PR #510, opened 2021-02-28
entities indexation: fix flattened terms query

maxlath:
Flattened terms currently don't help much for matching due to the different analyzer used during indexation (which you can see with the test added in d71208f3), this PR fixes that by adding a dedicated query with the query-time analyzer matching the index-time analyzer. I also created [a branch to track changes we would like to do before the next full reindexation of entities](https://github.com/inventaire/inventaire/tree/entities-reindexation), which would align the analyzers, allowing to remove that new query.

This PR also proposes to remove terms already in the main fields from the flattened terms, as those are just dead weight 
76f3896

### DIFF
--- a/server/db/elasticsearch/formatters/entity.js
+++ b/server/db/elasticsearch/formatters/entity.js
@@ -69,12 +69,23 @@ module.exports = async (entity, options = {}) => {
     entity.aliases = simplify.aliases(entity.aliases)
   }
 
+  const { labels, descriptions, aliases } = entity
+
+  entity.labels = removeUnusedLangs(labels)
   if (isWikidataEntity) {
-    // flattened terms are a single string to index every possible languages on a wikidata entity
-    // and to avoid having more than 1000 keys in an entity, which is forbidded by elasticsearch
-    entity.flattenedLabels = flattenTerms(entity.labels)
-    entity.flattenedDescriptions = flattenTerms(entity.descriptions)
-    entity.flattenedAliases = flattenTerms(entity.aliases)
+    entity.descriptions = removeUnusedLangs(descriptions)
+    entity.aliases = removeUnusedLangs(aliases)
+  }
+
+  const mainFieldsWords = new Set(getMainFieldsWords(entity))
+
+  if (isWikidataEntity) {
+    // Flattened terms are a single string to index all the words from non-active languages
+    // that weren't included in the fields above, as those non-active languages didn't get their
+    // own fields to avoid having more than 1000 keys in an entity, which is forbidded by elasticsearch
+    entity.flattenedLabels = flattenTerms(labels, mainFieldsWords)
+    entity.flattenedDescriptions = flattenTerms(descriptions, mainFieldsWords)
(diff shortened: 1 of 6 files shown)


tests/unit/indexation/entity_formatter.js
```diff
@@ -0,0 +1,34 @@
...
+require('should')
+const entityFormatter = __.require('db', 'elasticsearch/formatters/entity')
+
+describe('indexation: entity formatter', () => {
+  describe('flatten fields', () => {
+    it('should include words from inactive languages', async () => {
+      const doc = await entityFormatter(_.cloneDeep(Q535), { quick: true })
+      doc.flattenedLabels.includes('ᱦᱩᱜᱚ ᱵᱷᱤᱠᱴᱚᱨ').should.be.true()
+      doc.flattenedAliases.includes('Віктор-Мары').should.be.true()
+      doc.flattenedDescriptions.includes('թատերագիր').should.be.true()
```
jum-s, line 14: if we later open those languages in inventaire we would have to modify this test. I would suggest to only test if those `flattened*` are non empty strings

tests/api/search/entities.test.js
```diff
@@ -110,6 +110,13 @@ describe('search:entities', () => {
         _.map(results, 'uri').should.containEql(work.uri)
       })
     })
+
+    describe('not exact', () => {
+      it('should match flattened terms', async () => {
+        const results = await search({ types: 'series', search: 'ဟယ်ရီပေါ်တာ' })
```
jum-s, line 116: same as last comment what about finding the first key which is not in `activeI18nLangs` ?

server/db/elasticsearch/formatters/entity.js
```diff
@@ -132,10 +143,32 @@ const setTermsFromClaims = entity => {
...
-  return _.uniq(Object.values(terms)).join(' ')
+const flattenTerms = (terms, mainFieldsWords) => {
+  terms = Object.values(terms)
+  // Required for aliases
+  if (_.isArray(terms[0])) terms = _.flatten(terms)
+
+  return _.chain(terms)
+    .map(term => term.split(' '))
+    .flatten()
+    .filter(word => !mainFieldsWords.has(word.toLowerCase()))
```
jum-s, line 154: yep that make sense

tests/unit/indexation/entity_formatter.js
maxlath: that would strongly decrease the value of that test, to the point of not actually testing what the test label says. I would rather go with adding a fake language in the fixture, as we can safely assume that it will never be an active language: 18872b3b

tests/api/search/entities.test.js
maxlath: done in 1b4ec358

tests/unit/indexation/entity_formatter.js
jum-s: yeah it would decrease the relevance of this test, but i guess id rather have weaker tests than unstable ones.
this is non blocking

jum-s (approved):

PR #511, opened 2021-03-01
make entities history publicly accessible, while keeping contributors data private by default

maxlath:
Currently, entities history is only accessible by admins. This PR propose to open it to all, while keeping the data of the user that made the edit (a.k.a. contributor) private by default. The reasoning being that we want to preserve privacy by default, and that users might not know that when they add a subject to a work, they are actually editing a public wiki. This PR also allows to opt-in for public contributions by updating a setting on the user document. We could then recommend to regular contributors to opt-in, via the Wiki documentation and/or directly on the editor.

Client PR: #275

### DIFF
--- a/tests/api/entities/history.test.js
+++ b/tests/api/entities/history.test.js
@@ -1,9 +1,8 @@
-require('should')
+const should = require('should')
 const { adminReq, dataadminReq, publicReq, authReq, shouldNotBeCalled } = require('../utils/utils')
 const { createHuman } = require('../fixtures/entities')
 const { getDeanonymizedUser, customAuthReq } = require('../utils/utils')
 const endpoint = '/api/entities?action=history'
-const { _id: anonymizedId } = require('db/couchdb/hard_coded_documents').users.anonymized
 
 describe('entities:history', () => {
   it('should reject without uri', async () => {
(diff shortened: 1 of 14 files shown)


tests/api/entities/history.test.js
```diff
@@ -1,19 +1,21 @@
 require('should')
-const { adminReq, shouldNotBeCalled } = require('../utils/utils')
+const { adminReq, dataadminReq, publicReq, authReq, shouldNotBeCalled } = require('../utils/utils')
 const { createHuman } = require('../fixtures/entities')
+const { getDeanonymizedUser, customAuthReq } = require('../utils/utils')
 const endpoint = '/api/entities?action=history'
+const { _id: anonymizedId } = require('db/couchdb/hard_coded_documents').users.anonymized
```
jum-s, line 6: this list is supposed to be actual user doc in db. which is not the same as obfuscating an id
```suggestion
const anonymizedId  = 'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa'
```

maxlath: it's not only obfuscating, it being able to respond something when being requested on [`GET /api/users?action=by-ids`](https://inventaire.io/api/users?action=by-ids&ids=00000000000000000000000000000001), to lower the different cases to handle in clients

jum-s: why would there be any need to request such a user ? it lowers different cases maybe but if we could spare some requests, its kind of attractive no ?

tests/api/entities/history.test.js
```diff
@@ -22,7 +24,54 @@ describe('entities:history', () => {
...
+    patch.user.should.equal(anonymizedId)
+  })
+
+  it('should anonymize patches for authentified users', async () => {
+    const human = await createHuman()
+    const { patches } = await authReq('get', `${endpoint}&id=${human._id}`)
+    const patch = patches[0]
+    patch.user.should.be.a.String()
+    patch.user.should.equal(anonymizedId)
+  })
```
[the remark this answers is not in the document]
maxlath: I would rather make the expected behaviour explicit in the tests, and update the tests if we come to decide it should behave differently

tests/api/entities/history.test.js
maxlath: this is already expected by https://github.com/inventaire/inventaire-client/pull/275: if a non-admin user goes on the history page of an entity edited by anonymized users, patches will arrive with `user: ${anonymizedId}`, and make a `GET /api/users?action=by-ids` request for that id

jum-s: Why not deleting the `user` key in patches, and let clients fill what they cannot have access to ?

maxlath: that's a possible implementation, but we already have this existing mechanism to express special users, I find it cleaner to go with this implementation

maxlath: I tried without the special user in c8b258f1e, client update in https://github.com/inventaire/inventaire-client/pull/275/commits/ece525a2d23a1bdef07a2ec7155938da8a2d479d

jum-s (approved):

Issue #491, opened 2021-01-25
allow non-dataadmins users to merge entities


maxlath:
At the moment, only users with dataadmin rights can merge entities. I would be of the opinion that we should give non-dataadmin users the possibility to attempt a merge, while relying on some heuristic as for automatically accepting or declining that merge. Declined merges could then trigger the creation of a task to be considered by dataadmins

Elements to be considered in the heuristic:
* Do the labels of the merged entities match
* Do the labels of the entities in the merged entities graph match
* Would the merge only affect the requesting user's inventory items

jum-s:
could it be to simply trigger a `check_entity` which could then trigger a tasks creation or an automerge ?
We would still need to bring the `reqUserId` to `task.requester` and to the automerge (currently attributed to a `reconciler` user). But at least we can reuse the automerge heuristic without creating a new one.

jum-s:
Roadmap:

- [ ] open humans homonym deduplicates to non dataadmin users
- [ ] send and process server request to `check-human-duplicates` endpoint
- [ ] when creating task, adds `reporter` entry
- [ ] display reporter generated human tasks
- [ ] open works homonym deduplicates to non dataadmin users
- [ ] enable checking work entity fro duplicates

maxlath:
The problem I see with relying on `check_entity` automerge is for the case where the user requests a merge that is different from what the automerge would do: I think it make sense to let all users make requests to the `/api/entities?action=merge` endpoint (as started in #512), and then use  `check_entity` capabilities to check if the automerge agrees, and if not create a task

jum-s:
After some exploration, i reached the same conclusion, `/api/entities?action=merge` better fit the needs, also because check entity cannot accept a specific suggestion URI and will trigger multiple suggestion creation. It seem like a refactor to extract `automerge` from `checkEntity` is needed

jum-s:
About the heuristics [mentioned above](https://github.com/inventaire/inventaire/issues/491#issue-793505686), here is the current trend I'm leaning toward:

Merge if:
- One entity label AND at least one label of at least one graph entities match

Create task if:
- No entity label nor any graph entities label match

Note that those criteria are general to all "merge/create-task" possibilities, while some are specific to autogenerated tasks (`checkEntity`) aka too short works labels and work labels containing the author name are creating tasks.

Does this, a priori, make sense before I'm going too deep in the code?

jum-s:
Since #778, any logged-in user can ask for merging two entities of the same type.

Issue #91, opened 2017-05-19
allow users to suggest admin tasks, before being given admin rights


maxlath:
tasks that currently require admin access:

Already integrated to the client interface:
* merge entities

Not integrated to the client yet:
* change entity type
* change entity fixed entities relations (ex: [the work associated to an edition](https://github.com/inventaire/inventaire/issues/90)

Those tasks could be made visible to any user activating an "expert user" flag in user preferences, and after say 10 rightfully suggested task, this could unlock admin rights for this user.

jum-s:
Which mecanism grants a user new rights on the plateform (ie. merging entities) ?

- *by autority* : every an admin (or a priviledged user) can unlock new rights to users at their own discretion.
- *by review* : a user can suggest 10 (or x) 'to-merge' entities, then a reviewer can unlock the new right for this user.
- *by cooption* : when 2 (or x) priviledged users approve unlocking a right a user, it unlocks this right for the user.

My opinion is that by autority can lead to some [legacy problem](https://xkcd.com/1914/), and by review seems time intensive to code (create the 'to-merge' state, organise reviewing process, display users waiting list) while cooption can be a first step toward a full review process with minimal features (approve a right to one friend, approved users list for each user). 
I am curious for other possibilities of linkage mecanism

maxlath:
another option would be something like [MediaWiki `autoconfirmed` users](https://en.wikipedia.org/wiki/Wikipedia:User_access_levels#Autoconfirmed_and_confirmed_users), that is, users that automatically earn certain rights when meeting certain criteria. The right to merge could then be attributed to every users meeting those criteria, until proven not capable of not using this right correctly.

thewholelifetolearn:
Is there already a possibility to suggest the merge of entities through the client interface? I didn't find any button.

thewholelifetolearn:
Perhaps, before implementing an automatic system to "upgrade" an user, there should be a system to suggest actions to an admin.  
For example, in the same way as the "Feedback" form:
- add a "delete request" form on empty work or authors. The form should be pre-filled in with the page info. The user could just add a comment.
- add a "merge request" form for authors, works, editions, series. the form should be pre-filled in with the page info but the user must add the second entity to merge with. A comment can be added too.


Thereafter, an auto user upgrade system can be created.

maxlath:
@Alamantus https://github.com/inventaire/inventaire-client/issues/182
> ### Enable Merging Records by Users

[...]

PR #512, opened 2021-03-06
allow non-dataadmin to request merges

maxlath:
addressing #491 and #91 

by opening the `/api/entities?action=merge` endpoint to all authentified users. The current PR is limited to requesting merges by creating a task, but in the future, the merge could be done directly after some checks of what's at stake, just like it's done for `/api/entities?action=delete`

~WIP status: this PR depends depends on the `reqUserHasDataadminAccess` flag introduced in #513~

TODO:
- [ ] explore the possibility to use the task automerge (as suggested by https://github.com/inventaire/inventaire/issues/491#issuecomment-766925495)

### DIFF
--- a/server/controllers/entities/entities.js
+++ b/server/controllers/entities/entities.js
@@ -39,10 +39,10 @@ module.exports = {
       'update-label': require('./update_label'),
       'revert-edit': require('./revert_edit'),
       'restore-version': require('./restore_version'),
+      merge: require('./merge'),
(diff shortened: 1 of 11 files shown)


jum-s:
See if automerge, other entity type than works, work well out of the box (ie. `automerge` by `checkEntity`)

maxlath:
This need is still there but this PR is obsolete, let's give it a fresh start

PR #517, opened 2021-03-24
search: entities: fix exact match for terms containing single letters

maxlath:
example of a false negative match:
- [L'homme qui murmurait à l'oreille des chevaux](https://inventaire.io/api/search?types=works&search=L%27homme%20qui%20murmurait%20%C3%A0%20l%27oreille%20des%20chevaux&lang=fr&limit=100&exact=true) doesn't find wd:Q1448153
- [L'homme qui murmurait l'oreille des chevaux](https://inventaire.io/api/search?types=works&search=L%27homme%20qui%20murmurait%20%C3%A0%20l%27oreille%20des%20chevaux&lang=fr&limit=100&exact=true) does find wd:Q1448153

This can be fixed by making the autocomplete filter use `min_gram=1`: updating the prod Elasticsearch index settings won't be enough this time, a reindexation is required to fully benefit from the fix.

In case we go for a reindexation, we could also consider including the [commit on the `entities-reindexation` branch](https://github.com/inventaire/inventaire/compare/entities-reindexation)

### DIFF
--- a/server/db/elasticsearch/settings/settings.js
+++ b/server/db/elasticsearch/settings/settings.js
@@ -21,7 +21,7 @@ module.exports = {
       // See: https://www.elastic.co/guide/en/elasticsearch/reference/7.10/analysis-edgengram-tokenizer.html
       autocomplete_filter: {
         type: 'edge_ngram',
-        min_gram: 2,
+        min_gram: 1,
         max_gram: maxGram
       },
       // An analyzer to apply at search time to match the autocomplete analyzer used at index time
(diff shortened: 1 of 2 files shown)


jum-s (approved):

jum-s:
before accepting this, i would like to know the cost of this since this it greatly increasing the number of grams indexed

maxlath:
Another test case: https://inventaire.io/api/search?types=humans&search=Dennis%20E.%20Taylor&lang=en&exact=true should find `wd:Q41694161`

maxlath:
Current Elasticsearch operations are a bit chaotic (competing on RAM with the main server, keeping the RAM at 99%), so this PR could be let aside for now until we get more relaxed on ES ops

maxlath:
closing in favour of #608

Issue #481, opened 2021-01-05
Improve means of decentralization


jum-s:
moving toward more decentralization means to allow other people than the inventaire association to create inventaire instances (while having in mind issues such as [data replications](https://github.com/inventaire/inventaire/issues/187)). 
This issue intends to centralize issues about decentralization, which may involve (but is not limited to) : 
  - implement local basic books data import from bibliographical authority sources: https://github.com/inventaire/inventaire/pull/536
  - experiment how to reuse entities data from inventaire.io from another instance: https://github.com/inventaire/inventaire/pull/548
  - experiment how to edit inventaire.io entities data from another instance: https://github.com/inventaire/inventaire/issues/549

jum-s:
Review and improve documentation to setup a local inventaire server:
 - [x] review [docker-inventaire](https://github.com/inventaire/docker-inventaire) Readme (done with [743566](https://github.com/inventaire/docker-inventaire/commit/743566ced27da5a1b568c18415c216dc124861a0))
 - [ ] find some heuristic to simplify [`config/default.js`](https://github.com/inventaire/inventaire/blob/master/config/default.js) and publish the file
 - [ ] publish a simplified [nginx template](https://github.com/inventaire/inventaire-deploy/blob/master/nginx/inventaire.original.nginx)
 - [ ] add documentation about [inventaire-prerender](https://github.com/inventaire/prerender)
 - [ ] review and remove warning about decentralization in [inventaire-deploy Readme](https://github.com/inventaire/inventaire-deploy)

jum-s:
closing in favor of more granular label "decentralization"

PR #522, opened 2021-04-18
Convert image url locally

maxlath:
Currently, when adding an edition cover (`invp:P2` claim) from a URL, the URL is sent to the server, which sends it to dataseed, which does several things:
1 - cleanup the URL from known services (ex: it would convert https://covers.openlibrary.org/b/id/7811996-M.jpg into https://covers.openlibrary.org/b/id/7811996-L.jpg to get the larger image)
2 - download the remote image, generate a sha1 hash of the file, and re-upload it to the media storage container with that hash as filename
3 - add the source URL and corresponding hash to a CouchDB database

This PR proposes to move the points 2 and 3 to the inventaire server, to get the following benefits:
- increase independence from dataseed, contributing to #481, as instances without access to a dataseed service could still use the `/api/images?action=convert-url` endpoint
- simplify dataseed maintenance, by removing the need to administrate and backup an other CouchDB instance, and media storage logic
- generally reduces the complexity of how images are processed

client PR: https://github.com/inventaire/inventaire-client/pull/280

TODO after merge:
- import dataseed images database to inventaire new images database

### DIFF
--- a/server/controllers/images/lib/containers.js
+++ b/server/controllers/images/lib/containers.js
@@ -1,9 +1,7 @@
 const images_ = require('lib/images')
 const putImage = require('./put_image')
 
-const containerPutImage = (container, fnName) => fileData => {
-  const { id, path } = fileData
-
+const containerPutImage = (container, fnName) => ({ id = 0, path }) => {
   return images_[fnName](path)
   .then(() => images_.getHashFilename(path))
   .then(filename => putImage(container, path, id, filename))
--- a/server/controllers/images/lib/containers.js
+++ b/server/controllers/images/lib/containers.js
@@ -1,7 +1,8 @@
 const images_ = require('lib/images')
 const putImage = require('./put_image')
 
-const containerPutImage = (container, fnName) => ({ id = 0, path }) => {
+const containerPutImage = (container, fnName) => fileData => {
+  const { id = 0, path } = fileData
   return images_[fnName](path)
   .then(() => images_.getHashFilename(path))
   .then(filename => putImage(container, path, id, filename))
--- a/server/controllers/images/lib/import_image.js
+++ b/server/controllers/images/lib/import_image.js
@@ -1,0 +1,11 @@
+const { md5 } = require('lib/crypto')
+const downloadFile = require('./download_file')
+const { putImage } = require('controllers/images/lib/containers').entities
+
+module.exports = async sourceUrl => {
+  const tmpFilename = md5(sourceUrl)
+  const tmpPath = `/tmp/${tmpFilename}`
+
+  await downloadFile(sourceUrl, tmpPath)
+  return putImage({ path: tmpPath })
+}
--- a/tests/api/images/convert_url.test.js
+++ b/tests/api/images/convert_url.test.js
@@ -1,0 +1,12 @@
+require('should')
+const { authReq } = require('../utils/utils')
+const imageUrl = 'https://raw.githubusercontent.com/inventaire/inventaire-client/master/app/assets/icon/32.png'
+const endpoint = '/api/images?action=convert-url'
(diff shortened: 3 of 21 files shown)


server/controllers/images/lib/containers.js
```diff
@@ -1,9 +1,7 @@
 const images_ = require('lib/images')
 const putImage = require('./put_image')
 
-const containerPutImage = (container, fnName) => fileData => {
-  const { id, path } = fileData
-
+const containerPutImage = (container, fnName) => ({ id = 0, path }) => {
```
jum-s, line 4: not sure about those optimisation "au poil de cul", sometimes its useful to actually read what the object represent

maxlath: done in 7d341acde

server/controllers/images/lib/import_image.js
```diff
@@ -0,0 +1,11 @@
+const { md5 } = require('lib/crypto')
+const downloadFile = require('./download_file')
+const { putImage } = require('controllers/images/lib/containers').entities
+
+module.exports = async sourceUrl => {
+  const tmpFilename = md5(sourceUrl)
```
jum-s, line 6: md5 ? why not simply base64

tests/api/images/convert_url.test.js
```diff
@@ -0,0 +1,37 @@
...
+
+describe('images:convert-url', () => {
+  it('should convert a URL', async () => {
+    const imageUrl = 'https://raw.githubusercontent.com/inventaire/inventaire-client/master/app/assets/icon/32.png'
+    const { url, hash } = await convertUrl(imageUrl)
+    isImageHash(hash).should.be.true()
+    isLocalImg(url).should.be.true()
+  })
+
+  it('should reject an URL returning a 404', async () => {
```
jum-s, line 17: maybe a test on rejecting non URL string?

tests/api/images/upload.test.js
```diff
@@ -0,0 +1,47 @@
...
+describe('images:upload', () => {
+  it('reject uploads on assets containers', async () => {
+    await authReq('post', `${endpoint}&container=assets`)
+    .then(shouldNotBeCalled)
+    .catch(err => {
+      err.body.status_verbose.should.startWith('invalid container')
+    })
+  })
+
+  it('reject uploads on remote containers', async () => {
```
jum-s, line 23: i dont understand the purpose of those 2 `reject` tests (assets & remote) if non of those parameters are in a `should` test later. Does that mean its possible to post a `container=remote` ? If yes would be nice to document it with a test, no?

tests/api/images/convert_url.test.js
maxlath: done in 5dcfc4625

tests/api/images/upload.test.js
maxlath: It's not possible to post a `container=remote`, no, only on `entities` and `users` containers, but `assets` is a container (where uploads are done directly by an instance admin, without passing by this endpoint) and `remote` is a pseudo-container, known by the `/img` endpoint, but that doesn't have a container in the object storage (source image files aren't fetched from the object storage but from the remote URL). Hopefully that's clearer, do you still see a missing test?

jum-s: it is clearer, but i still dont get why we test 2 rejections without any acceptation tests later. if i push the logic why not testing that `container=foo` only and not twice like now ?
In any case could we paste your comment inside the test file as a documentation, to not lose it in the limbs of gh comments?

maxlath: acceptance tests added in a7d6c3c20 and documentation added in 7ec56c963

jum-s (approved):

PR #530, opened 2021-05-18
search claim filter

maxlath:
This PR proposes a first implementation of search claim filters, inspired by [Wikibase `haswbstatement`](https://www.mediawiki.org/wiki/Help:Extension:WikibaseCirrusSearch#haswbstatement).

Implemented:
* [x] `AND`
* [x] `OR`
~~* [ ] `NOT`~~ (maybe it's fine to let that one out from this PR?)

Those filters would allow, for instance, to add search fields on authors layouts: in that example, the `claim` parameter would be set to `wdt:P50=${author.uri}` (or actually more probably `wdt:P50=${author.uri}|wdt:P58=${author.uri}|wdt:P110=${author.uri}|wdt:P6338=${author.uri}` to be complete)

This branch is based on `213-relations-search`, so most commits displayed below at the time of this message actually belong to #529, which should thus be considered first

### DIFF
--- a/server/controllers/search/lib/entities_query_builder.js
+++ b/server/controllers/search/lib/entities_query_builder.js
@@ -109,7 +109,7 @@ const entitiesFields = (userLang, exact) => {
 const getClaimFilter = claimParameter => {
   return {
     terms: {
-      claim: [ claimParameter ]
+      claim: claimParameter.split('|')
     }
   }
 }
--- a/server/controllers/search/lib/entities_query_builder.js
+++ b/server/controllers/search/lib/entities_query_builder.js
@@ -14,7 +14,7 @@ module.exports = params => {
     { terms: { type: types } }
   ]
 
-  if (claim) filters.push(...getClaimFilter(claim))
+  if (claim) filters.push(...getClaimFilters(claim))
 
   if (!search) minScore = 0
 
@@ -109,7 +109,7 @@ const entitiesFields = (userLang, exact) => {
   return fields
 }
 
-const getClaimFilter = claimParameter => {
+const getClaimFilters = claimParameter => {
   return claimParameter
   .split(' ')
   .map(andCondition => {
--- a/server/controllers/search/lib/entities_query_builder.js
+++ b/server/controllers/search/lib/entities_query_builder.js
@@ -18,19 +18,18 @@ module.exports = params => {
 
   if (!search) minScore = 0
 
+  const shoulds = matchEntities(search, userLang, exact)
+
   return {
     query: {
       function_score: {
         query: {
           bool: {
             filter: filters,
-            // Because most of the work has been done at index time (indexing terms by ngrams)
-            // all this query needs to do is to look up search terms which is way more efficient than the match_phrase_prefix approach
-            // See https://www.elastic.co/guide/en/elasticsearch/guide/current/_index_time_search_as_you_type.html
-            should: matchEntities(search, userLang, exact),
-            // The default value would be 0 due to the presence of a filter
+            should: shoulds,
+            // The default value would be 0 due to the presence of filters
             // See https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-bool-query.html#bool-min-should-match
-            minimum_should_match: 1
+            minimum_should_match: shoulds != null ? 1 : 0
           }
         },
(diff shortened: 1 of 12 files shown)


jum-s:
how far is a feature `author=${author.uri} illustrator=${author.uri} senarist=${author.uri} colorist=${author.uri}` (note the space separated terms)
step 1: in english?
step 2: is the user language?

maxlath:
> how far is a feature author=${author.uri} illustrator=${author.uri} senarist=${author.uri} colorist=${author.uri} 

That could probably be done already in the current state of this PR, by having the client substitute `"author"` by `wdt:P50`, etc. The client could then also consider i18n terms and also substitute, for instance, `"auteur"` with `wdt:P50` when the user language is `fr`

server/db/elasticsearch/formatters/entity.js
```diff
@@ -171,3 +176,41 @@ const getMainFieldsWords = ({ labels, descriptions = {}, aliases = {} }) => {
...
+  'wdt:P179', // serie
+  'wdt:P6338', // colorist
+]
+
+const indexedRelationsPerType = {
+  collection: [
+    'wdt:P123', // publisher
+  ],
+  serie: worksAndSeriesProperties,
+  work: worksAndSeriesProperties,
```
jum-s, line 201: "a serie is a work like another" shown by code

jum-s:
todo after merge: update api.inventaire.io

server/controllers/search/lib/entities_query_builder.js
```diff
@@ -96,15 +95,47 @@ const entitiesFields = (userLang, exact) => {
...
-const entitiesFlattenedFields = [
-  'flattenedLabels^0.25',
-  'flattenedAliases^0.25',
-  'flattenedDescriptions^0.1'
-]
+const getClaimFilters = claimParameter => {
+  return claimParameter
+  .split(' ')
+  .map(andCondition => {
+    const orConditions = andCondition.split('|')
```
jum-s, line 113: what about stripping spaces in case of typo, in case of messy client?
```suggestion
    const orConditions = andCondition.split(/\D\|\D/)
```
or maybe not worth the regex cost
```suggestion
    const orConditions = andCondition.split('|').map(orCondition => orCondition.trim())
```

tests/api/search/entities.test.js
```diff
@@ -142,6 +141,60 @@ describe('search:entities', () => {
...
+        const results = await search({ types: 'works', claim: `wdt:P31=wd:Q47461344 wdt:P50=${workAuthor.uri}`, lang: 'en', filter: 'inv' })
+        const foundIds = _.map(results, 'id')
+        foundIds.should.containEql(workWithAuthor._id)
+        const results2 = await search({ types: 'works', claim: `wdt:P31=wd:Q2831984 wdt:P50=${workAuthor.uri}`, lang: 'en', filter: 'inv' })
+        const foundIds2 = _.map(results2, 'id')
+        foundIds2.should.not.containEql(workWithAuthor._id)
+      })
+
+      it('should accept a combination of AND and OR conditions', async () => {
+        const results = await search({ types: 'works', claim: `wdt:P31=wd:Q47461344 wdt:P50=wd:Q535|wdt:P50=${workAuthor.uri}`, lang: 'en', filter: 'inv' })
```
jum-s, line 193: now that development is over, what about moving describe claim block to its own file, and probably replace `wd:Q47461344` by a dynamic value inv uri instead of relying on external uri?

tests/api/search/entities.test.js
```diff
@@ -142,6 +141,60 @@ describe('search:entities', () => {
...
+      })
+
+      it('should find an entity by one of its relation claims', async () => {
+        const results = await search({ types: 'works', claim: `wdt:P50=${workAuthor.uri}`, lang: 'en', filter: 'inv' })
+        const foundIds = _.map(results, 'id')
+        foundIds.should.containEql(workWithAuthor._id)
+      })
+
+      it('should accept OR conditions', async () => {
+        const results = await search({ types: 'works', claim: `wdt:P50=wd:Q535|wdt:P50=${workAuthor.uri}`, lang: 'en', filter: 'inv' })
```
jum-s, line 178: now that development is over, what about (independent possibilities): 
- moving describe claim block to its own file
- replace `wd:Q535` by a dynamic value 
- have an inv uri instead of relying on external uri

jum-s:
what about having both `search` and `claims` ?

PR #535, opened 2021-05-28
controllers deboilerplatification

maxlath:
This PR proposes to move controllers boilerplate to [`server/lib/actions_controllers.js`](https://github.com/inventaire/inventaire/blob/a930c1c/server/lib/actions_controllers.js), taking `server/controllers/shelves/update.js` as an example of what could be done (leaving the application in bulk of this new format to another PR)


jum-s (approved):
looking good to me

PR #536, opened 2021-06-11
autocreate entities from authorities-sourced entries

maxlath:
This PR proposes to make use of national libraries open data (aka authorities) to build entries that can be resolved using the resolver, and create the corresponding missing entities when not resolved. `owl:sameAs` triples come to be particularly valuable, as those can then be used to increase chances to resolve to an existing Wikidata entity.

The 3 authorities queried in this PR were just picked to be able to test requesting several databases at once and come up with a way to determine what should be passed to the resolver in the case where several databases return an entry. More authorities could be added later:
* [KB](http://libris.kb.se/sparql)
* GND via the [unofficial SPARQL endpoint](https://zbw.eu/beta/sparql-lab/?endpoint=http://zbw.eu/beta/sparql/gnd/query)
* [LoC data](https://id.loc.gov/) ([example](http://lx2.loc.gov:210/lcdb?version=1.1&operation=searchRetrieve&query=bath.isbn=9783540741190&maximumRecords=1&recordSchema=dc))
* probably more


server/data/lib/external_ids.js
```diff
@@ -0,0 +1,70 @@
+const { uniq, flatten } = require('lodash')
+const getEntityIdBySitelink = require('data/wikidata/get_entity_id_by_sitelink')
+const properties = require('controllers/entities/lib/properties/properties_values_constraints')
+
+// Accepts several string arguments, either as single URLs or as a group of urls concatenated with ',' as separator
+const parseSameAsMatches = async (...matches) => {
```
jum-s, line 6:
```suggestion
const parseSameasMatches = async (...matches) => {
```

maxlath: done in f202901d6

server/data/dataseed/get_authorities_aggregated_entry.js
```diff
@@ -0,0 +1,60 @@
...
+  bne: require('data/bne/get_bne_entry_from_isbn'),
+  bnf: require('data/bnf/get_bnf_entry_from_isbn'),
+  wikidata: require('data/wikidata/get_wikidata_entry_from_isbn'),
+}
+
+const authoritiesNames = Object.keys(authorities)
+
+module.exports = async isbn => {
+  const entries = await Promise.all(authoritiesNames.map(wrap(isbn)))
+  // TODO: aggregate resolved data, rather than just returning the best one
```
jum-s, line 16: would be nice to actually to this todo before merge. We might miss some great data because of this ?
do you mind if i try @maxlath ?

maxlath: @jum-s you're welcome to try, but if the attempt isn't that succesful, I would prefer not having it blocking this PR as that should already be in improvement compared to the current situation

jum-s: i know, but all that PR to discard claims not found in the best authority is quite sad

jum-s: done with `5a7cf9ea` (branch has been rebased too)

PR #538, opened 2021-06-22
Activity pub

jum-s:
Implement 
- webfinger actor discoverability (by username)
- signed request for actor and inbox endpoints
- some activity stream "Follow" example, which create an activity if a POST on ?action=inbox&name{:username} is received
- possibility to optin in user settings to make user `discoverable` (undiscoverable user respond a 403)


maxlath (changes requested):
first part of the review: I still have to look at the tests

server/db/couchdb/design_docs/activities.json
```diff
@@ -0,0 +1,9 @@
+{
+  "_id": "_design/activities",
+  "language": "coffeescript",
+  "views": {
+    "byExternalIds": {
```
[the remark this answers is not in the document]
maxlath: addressed in 94fbeb6f6

server/lib/actions_controllers.js
```diff
@@ -38,14 +39,15 @@ module.exports = controllers => {
     try {
       if (sanitization) {
         const params = sanitize(req, res, sanitization)
+        await verifySignedReq(req, res)
```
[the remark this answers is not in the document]
maxlath: Alternatively, the call to `verifySignature` could be moved to controllers themselves, like controllers that throw `unauthorizedApiAccess` errors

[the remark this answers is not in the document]
jum-s: after playing with the outbox, its nice to regroup all signature request validation in one place, but it looks like controllers can handle it themselves

tests/api/activitypub/actor.test.js
```diff
@@ -0,0 +1,56 @@
...
+
+describe('activitypub:actor', () => {
+  it('should reject unknown actor', async () => {
+    try {
+      const emitterUser = await createUserOnFediverse()
+      const { origin, query } = await startServerWithEmitterUser({ emitterUser })
+      const keyUrl = makeUrl({ origin, params: query })
+      const imaginaryReceiverUsername = createUsername()
+      const receiverUrl = makeUrl({ params: { action: 'actor', name: imaginaryReceiverUsername } })
+      await signedReq({ url: receiverUrl, keyUrl, privateKey: emitterUser.privateKey })
```
maxlath, line 14: The internals of `signedReq` (which user from what server are doing the signed request) aren't so relevant to those tests, it would be nice if `signedReq` could hide those (like for instance`adminReq` is doing) by keeping a fediverse server and user at hand, and default to those when not told otherwise
```suggestion
      const imaginaryReceiverUsername = createUsername()
      const receiverUrl = makeUrl({ params: { action: 'actor', name: imaginaryReceiverUsername } })
      await signedReq({ url: receiverUrl })
```

tests/api/activitypub/signed_request.test.js
```diff
@@ -0,0 +1,160 @@
...
+      rethrowShouldNotBeCalledErrors(err)
+      const parsedBody = JSON.parse(err.body)
+      parsedBody.status.should.equal(400)
+      parsedBody.status_verbose.should.equal('no signature header')
+    }
+  })
+
+  it('should reject when no publicKey is found', async () => {
+    try {
+      const emitterUser = await createUserOnFediverse()
```
maxlath, line 48: It's a bit confusing that the `emitterUser` supposedly from another instance, is actually created in the local instance database. It would be clearer if `tests/api/utils/activitypub.js` was just using directly `createKeyPair`, relying on simply in-memory user data

tests/api/utils/activitypub.js
```diff
@@ -0,0 +1,79 @@
...
+  const host = `localhost:${port}`
+  const origin = `http://${host}`
+  const webfingerEndpoint = '/.well-known/webfinger?resource='
+  const resource = `acct:${username}@${host}`
+
+  for (const endpt of endpoints) {
+    app.get('/' + endpt.name, (req, res) => {
+      return res.json(endpt.resData)
+    })
+  }
```
maxlath, line 52: yagni indeed
```suggestion
```

tests/api/utils/activitypub.js
```diff
@@ -0,0 +1,79 @@
...
+  app.get(`${webfingerEndpoint}${resource}`, async (req, res) => {
+    return res.json(formatWebfinger(origin, endpoint, resource))
```
maxlath, line 54: the query string is ignored (you can verify that assertion by replacing `resource=` by `rezzzzzzzzzz=`, all tests pass)
```suggestion
  app.get('/.well-known/webfinger', async (req, res) => {
    const { resource } = req.query
    return res.json(formatWebfinger(origin, endpoint, resource))
```

tests/api/utils/activitypub.js
```diff
@@ -0,0 +1,79 @@
...
+const startServerWithEmitterUser = async ({ emitterUser }) => {
+  const { origin } = await startActivityPubServer({ user: emitterUser })
+  return {
+    origin,
+    query: { action: 'actor', name: emitterUser.username }
+  }
+}
+
+const startActivityPubServer = ({ user, endpoints = [] }) => new Promise(resolve => {
+  const port = 1024 + Math.trunc(Math.random() * 10000)
```
maxlath, line 40: wouldn't a single server be enough? do we need a different server for each of those tests?

tests/api/utils/request.js
```diff
@@ -99,12 +100,38 @@ const bearerTokenReq = (token, method, endpoint, body) => {
   })
 }
 
+const signedReq = async ({ method, endpoint, url, keyUrl, privateKey, body }) => {
```
maxlath, line 103: As suggested above in the review of the `actor` tests, most consumers of `signedReq` just want to make a request from some valid fediverse user, so having default values for `keyUrl` and `privateKey` would be convenient
```suggestion
let someFediverseUser
const signedReq = async ({ method, endpoint, url, keyUrl, privateKey, body }) => {
  if (!keyUrl) {
    someFediverseUser = someFediverseUser || await createUserOnFediverse()
    keyUrl = someFediverseUser.keyUrl
    privateKey = someFediverseUser.privateKey
  }
```

maxlath (changes requested):
1st review, 2nd part: tests

server/lib/actions_controllers.js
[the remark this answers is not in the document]
maxlath: what about having the, now on master, `controllerWrapper` function accept `access=signed`, which would allow to finely set which endpoint should be signed:
```js
// server/controllers/activitypub/activitypub.js could look like this
module.exports = {
  get: ActionsControllers({
    public: {
      // Btw, that endpoint actually doesn't really need to be signed, does it?
      // It doesn't write anything and only returns public data, right?
      actor: require('./actor')
    }
  }),
  post: ActionsControllers({
    signed: {
      inbox: require('./inbox')
    }
  })
}
```

tests/api/utils/activitypub.js
jum-s: i would say whatever the test server respond to, its not inventaire server problem
i added a test anyway and apply the suggestion above in 9f18b486

tests/api/utils/activitypub.js
maxlath: done in 465a36c

maxlath (approved):

PR #513, opened 2021-03-06
allow to make entities contributions public

maxlath:
this PR depends on #511, and in itself contains only 1 commit (4b7101ca), the others belonging to #511

PR #539, opened 2021-06-30
merge activity endpoint in contributions controller

jum-s:
the idea is to get rid of the confustion between a contributions
activity and an activity-stream activity

it makes the choice of having a more complex contributions controller,
but the semantics makes more sense (an activity was actually a
contribution)


maxlath:
One controller expects a `user`, `limit`, `offset`, the other a `period`: ideally, undesired parameters should trigger a warning in the response, so merging both would mean having to do that manually, instead of letting the sanitization do its job. Additionally, the access level for those 2 controllers can be different (that would be the case if #513 is merged), meaning having to do  access validation in the controller. What about simply renaming the `/api/entities?action=activity` endpoint to something like `/api/entities?action=contributions-count`?

jum-s:
done in 604b8422a25d85cfc25897d4bc584672a9de2c39

maxlath (approved):
can be merged once https://github.com/inventaire/inventaire-client/pull/287 is updated and mergeable

PR #540, opened 2021-07-09
controllers: extract non-action-related logic from ActionsControllers function

maxlath:
* allows to declare endpoints in a deboilerplatified way, even when not passing by `ActionsControllers`

Bonus:
* remove last bits of `sanitizeAsync`

### DIFF
--- a/server/lib/actions_controllers.js
+++ b/server/lib/actions_controllers.js
@@ -22,13 +22,11 @@ module.exports = controllers => {
   return async (req, res) => {
     // Accepting the action to be passed either as a query string
     // or as a body parameter for more flexibility
-    let action = req.query.action || req.body.action
+    const action = req.query.action || req.body.action || 'default'
 
-    if (action == null && actionsControllersParams.default == null) {
+    if (action === 'default' && actionsControllersParams.default == null) {
       return error_.bundleMissingQuery(req, res, 'action')
     }
-
-    action = action || 'default'
 
     const controllerParams = actionsControllersParams[action]
     if (controllerParams == null) return error_.unknownAction(req, res)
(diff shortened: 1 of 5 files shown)


server/lib/actions_controllers.js
```diff
@@ -1,87 +1,68 @@
...
-    if (action == null && actions.default == null) {
+    if (action == null && actionsControllersParams.default == null) {
       return error_.bundleMissingQuery(req, res, 'action')
     }
 
-    const actionData = action ? actions[action] : actions.default
-    if (actionData == null) {
-      return error_.unknownAction(req, res)
-    }
+    action = action || 'default'
```
jum-s, line 31: e

server/lib/actions_controllers.js
```diff
@@ -1,87 +1,68 @@
...
-  const actions = getActions(controllers)
+  const actionsControllersParams = getActionsControllersParams(controllers)
   return async (req, res) => {
     // Accepting the action to be passed either as a query string
     // or as a body parameter for more flexibility
-    const action = req.query.action || req.body.action
+    let action = req.query.action || req.body.action
 
-    if (action == null && actions.default == null) {
+    if (action == null && actionsControllersParams.default == null) {
```
jum-s, line 27: strickly(?) equivalent suggestion (that cannot fit inside suggestion commenting tool)

```
    const action = req.query.action || req.body.action || 'default'

    if (action === 'default' && actionsControllersParams.default == null) {
```

and remove `action = action || 'default'` below

maxlath: refactor done in f26e02c67

Issue #301, opened 2019-05-15
CORS: open the API to JS requests in browsers


mytskine:
I tried to use the API from the JS of a web page, and I failed. I may be wrong, but I think that the API is not open to this kind of requests.

Here is a typical response from the API to a first request:
```
% curl -i -X OPTIONS \
  -H 'Access-Control-Request-Method: POST' \
  -H 'Access-Control-Request-Headers: content-type' \
  'https://inventaire.io/api/auth?action=login'
HTTP/2 400 
[...]
access-control-allow-origin: https://api.inventaire.io
access-control-allow-methods: GET,POST,PUT
access-control-allow-credentials: true
```

So the response header [access-control-allow-origin](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin) says that only a browser visiting `https://api.inventaire.io` is allowed to consume this API. Since there is no `Vary` header, it means this restriction will not vary with the source of the request, so api.inventaire.io is definitely the only source allowed here.

Another problem is that the method "OPTIONS" is rejected (HTTP 400). But if the browser wants to send a POST request with a JSON payload to an external server, it must send an OPTIONS request first, to ask if this combination of method and content-type is allowed. This is mandatory for [preflight requests](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS#Preflighted_requests), as specified by CORS (Cross-Origin Resource Sharing).

maxlath:
Hi, sorry for the late answer ^^'
I tried to address this issue in #543. Note that the proposed change would make the above `OPTIONS` request receive a 200 response with `access-control-allow-origin: *`, `access-control-allow-methods: *`, but no `access-control-allow-credentials`, so the login request won't be able to return session cookies

PR #543, opened 2021-08-14
CORS policy update

maxlath:
Addressing #301

* accept preflight requests by always returning a 200 with `access-control-allow-*` headers to `OPTIONS` requests
* open all public resources to cross-domain request
* deny access to resources requiring authentication by advising browsers to not include session cookies in cross-domain requests (by setting `access-control-allow-credentials=false`), to prevent CSRF

Authenticated requests from third party apps should then rather pass by OAuth tokens #84

### DIFF
--- a/server/lib/requests.js
+++ b/server/lib/requests.js
@@ -167,5 +167,6 @@ module.exports = {
   put: req('put'),
   delete: req('delete'),
   head,
+  options: req('options'),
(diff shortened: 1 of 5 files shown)


jum-s (approved):
approving to not block this PR, but i dont understand the subtleties written here, its beyond my current understanding.

Issue #213, opened 2018-09-30
make search find books by title AND author


maxlath:
ex: "stefan zweig montaigne" should find https://inventaire.io/entity/inv:5ff7506a618ed1f61bcf6d46400e40c0

that would require to index authors labels in works entries

jum-s:
index authors labels in works entries seems to be the responsibility of the entity-search-engine when posting to elastic new entities and updated entities. This would need an update everytime some author-work relationship is added/deleted and when authors name are modified too. It would duplicate data of labels stored in another index.

But Im wondering if a combination of [aliases](https://www.elastic.co/guide/en/elasticsearch/reference/6.8/alias.html) (duplicating the field name labels of works with  `work_labels`) and a [join datatype](https://www.elastic.co/guide/en/elasticsearch/reference/6.8/parent-join.html) (since its in the same index `entities`) could do the trick.

[documentation](https://www.elastic.co/guide/en/elasticsearch/reference/5.6/parent-join.html#_parent_join_and_performance) notes that the join datatype "adds a significant tax to your query performance" and is only available from de 5.6 version.

jum-s:
[a conversation with max] The join datatype does not seem to fit since relation can only exist within the same index (since `wikidata` and `entities` indexes have relations, ie. a wikidata author with a inventaire work).
Best effort feature : when i search for a work and results are not satisfying, then i can add the author name to have a more precise query (ex: typing `the ballad` then `Susan Andersen`). 
It could make sense to index authors labels and statements within works documents when [formatting the entity](https://github.com/inventaire/inventaire/blob/master/server/db/elasticsearch/formatters/entity.js)

maxlath:
done in #529: [searching for a work with "Albert Findling" returns his work "Du mot à la phrase"](https://inventaire.io/api/search?types=works&search=Albert%20Findling&lang=fr&limit=10&filter=inv) :tada:

PR #545, opened 2021-09-08
search: recover from cross_fields search errors

maxlath:
`cross_fields` matching was disabled by 99a28ab02 to prevent the occasional crashes (see https://github.com/elastic/elasticsearch/issues/44700), but is actually critical to resolving https://github.com/inventaire/inventaire/issues/213 thus this compromise: default to `cross_fields`, but if the request to Elasticsearch fails, fall back to `best_fields`


jum-s:
case of query that fails with `cross_fields`

[https://inventaire.io/search?q=mechanical+kelly](https://inventaire.io/search?q=mechanical+kelly)

jum-s (approved):

PR #548, opened 2021-09-15
experimentation on how to reuse entities data from another instance

maxlath:
Reusing the entities from another instance seems to be possible. What this does:
* disable local entities/patches/tasks databases and controllers
* for pieces of code from other controllers that were trying to access those databases (ex: to generate items snapshots), adds an abstraction layer: instance-agnostic functions, such as `getEntitiesByUris`, where entities are then fetched either locally or remotely depending on the configuration
* Search on entities was disable: clients are expected to perform entities search request on the remote instance instead

Other things that could be done:
* The current implementation doesn't do caching, so any call to `getEntitiesByUris` means doing requests to the remote instance
* Wikidata entities are also fetched on the remote Inventaire instance: directly fetching Wikidata entities on wikidata.org could give more autonomy from the remote Inventaire instance

Identified limitations:
* Some event hooks usually triggered by the local entities code (ex: `entity:update:label`, `entity:update:claim`, `entity:merge`, `entity:revert:merge`) will ~~never be triggered~~ not be triggered each time an entity from the remote instance is updated, as there is currently no update channel to listen to, leading to cached data not being invalidated (ex: again, items snapshots). Those events could be triggered when the update comes from the local instance though (as #549 would allow), just like for updates on Wikidata entities.

Client side branch: https://github.com/inventaire/inventaire-client/pull/291

### DIFF
--- a/config/default.js
+++ b/config/default.js
@@ -28,6 +28,8 @@ const config = module.exports = {
     return `${this.publicProtocol}://${this.publicHost}:${this.port}`
   },
   invHost: 'https://inventaire.io',
+  // Set to 'https://inventaire.io' in ./local-dev.js to use inventaire.io entities locally
+  remoteEntities: null,
   // To allow fallback between servers, they need to share the same session keys:
   // one should have autoRotateKeys=true and the others autoRotateKeys=false
   autoRotateKeys: true,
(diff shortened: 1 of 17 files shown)


maxlath:
Closing this PR for now, as it was an experimentation and additional work would be needed to make it a sustainable approach (see comment above), but it could be re-opened later

maxlath:
It might actually be interesting to have the local server proxy the remote entities requests as that would have the following benefits:
- the local server could use those requests redirections to trigger hooks such as `entity:merge` (but not `entity:revert:merge`(?))
- better privacy, as entities read requests would not need to be disclosed to the remote server

PR #566, opened 2021-11-30
sanitization: parameter aliases and config validation at start-up

maxlath:
Does 2 features were coded together but do not depend on each others (could be splitted, if one is blocking)

## parameter aliases
Allow to use a well known sanitization parameter (ex: `password`) with a prefix (ex: `new`) without having to set yet another parameter in `server/lib/sanitize/parameters.js`

## config validation at start-up
Until this, if a controller was trying to use parameter not defined in `server/lib/sanitize/parameters.js`, that bug would only be raised when a user would make a request.

### DIFF
--- a/server/controllers/auth/connection.js
+++ b/server/controllers/auth/connection.js
@@ -1,14 +1,14 @@
-const { sanitize } = require('lib/sanitize/sanitize')
+const { sanitize, validateSanitization } = require('lib/sanitize/sanitize')
 const error_ = require('lib/error/error')
 const passport_ = require('lib/passport/passport')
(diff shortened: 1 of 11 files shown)


jum-s (approved):

jum-s:
of course i just accept the PR and saw afterward on another PR the extension of the regex `prefixedParameterPattern` from `/^(old|new)-/` to `/^(old|new|current)-/`, and saw the extensibility ad infinitum. I would suggest to not include `-` at the end of the pattern, so that any prefix could then be added, not necessarly with a dash. Im saying this now since our future selves might be afraid to break things by removing the dash in the future

suggestion: 
`const prefixedParameterPattern = /^(old-|new-)/`

PR #586, opened 2021-12-12
allow to filter contributions by claim property or label lang

maxlath:
A useful tool for administrators to misinformed edits on a precise property: my current use-case is a user having change many `wdt:P629`, and those would need to be reverted, but it needs to be done manually as some entities were further edited.

This PR depends on #585 for convenience, but could be rewritten without it if #585 was to get stuck

### DIFF
--- a/server/controllers/entities/contributions.js
+++ b/server/controllers/entities/contributions.js
@@ -7,7 +7,11 @@ const sanitization = {
   limit: { default: 100, max: 1000 },
   offset: { default: 0 },
   property: { optional: true },
-  lang: { optional: true },
+  lang: {
+    optional: true,
+    // Override global parameter default
+    default: null,
+  },
 }
 
 const controller = params => {
--- a/server/controllers/entities/contributions.js
+++ b/server/controllers/entities/contributions.js
@@ -1,17 +1,22 @@
 // An endpoint to list entities edits made by a user
-const patches_ = require('./lib/patches')
+const { byUserId, byDate, byUserIdAndProperty } = require('./lib/patches')
 
 const sanitization = {
   user: { optional: true },
   limit: { default: 100, max: 1000 },
-  offset: { default: 0 }
+  offset: { default: 0 },
+  property: { optional: true },
 }
 
-const controller = ({ userId, limit, offset }) => {
+const controller = ({ userId, limit, offset, property }) => {
   if (userId != null) {
-    return patches_.byUserId(userId, limit, offset)
+    if (property) {
+      return byUserIdAndProperty({ userId, property, limit, offset })
+    } else {
+      return byUserId({ userId, limit, offset })
+    }
   } else {
-    return patches_.byDate(limit, offset)
+    return byDate({ limit, offset })
   }
 }
 
--- a/server/controllers/entities/contributions.js
+++ b/server/controllers/entities/contributions.js
@@ -1,29 +1,29 @@
 // An endpoint to list entities edits made by a user
 const { byUserId, byDate, byUserIdAndFilter } = require('./lib/patches')
 const error_ = require('lib/error/error')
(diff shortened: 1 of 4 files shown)


server/controllers/entities/contributions.js
```diff
@@ -1,17 +1,22 @@
 // An endpoint to list entities edits made by a user
-const patches_ = require('./lib/patches')
+const { byUserId, byDate, byUserIdAndProperty } = require('./lib/patches')
 
 const sanitization = {
   user: { optional: true },
   limit: { default: 100, max: 1000 },
-  offset: { default: 0 }
+  offset: { default: 0 },
+  property: { optional: true },
```
jum-s, line 9: the couch view comment `property can be both a label lang or a claim property`, below, could be duplicated (or moved) here, to know early that property could also be label

tests/api/entities/contributions.test.js
```diff
@@ -76,6 +77,24 @@ describe('entities:contributions', () => {
...
+
+  describe('filter', () => {
+    it('should filter by claim property', async () => {
+      const user = await getReservedUser()
+      const work = await createWork({ user })
+      const property = 'wdt:P921'
+      await updateClaim({ uri: work.uri, property, newValue: 'wd:Q1', user })
+      await updateClaim({ uri: work.uri, property: 'wdt:P136', newValue: 'wd:Q2', user })
+      await updateClaim({ uri: work.uri, property, newValue: 'wd:Q3', user })
+      await updateClaim({ uri: work.uri, property, oldValue: 'wd:Q1', user })
```
jum-s: another way to explicitly say "property could be labels" (cf last comment) could be to add a test also

server/controllers/entities/contributions.js
```diff
@@ -1,14 +1,24 @@
...
 
-const controller = ({ userId, limit, offset, property }) => {
+const controller = params => {
+  const { userId, limit, offset, lang } = params
+  let { property } = params
+  if (property && lang) {
+    throw error_.new('ca, not use both property and lang filters', 400, params)
+  }
+
+  property = property || lang
```
jum-s, line 20: (cf last comment) maybe this line is sefl documenting enough

server/controllers/entities/contributions.js
```diff
@@ -1,17 +1,32 @@
...
+  property: { optional: true },
+  lang: { optional: true },
 }
 
-const controller = ({ userId, limit, offset }) => {
+const controller = params => {
+  const { userId, limit, offset, lang } = params
+  let { property } = params
+  if (property && lang) {
+    throw error_.new('ca, not use both property and lang filters', 400, params)
```
jum-s, line 17: typo

```suggestion
    throw error_.new('can not use both property and lang filters', 400, params)
```

server/controllers/entities/contributions.js
maxlath: In the current PR implementation, there are 2 possible parameters that get a tailored sanitization: passing a `lang` value as `property` would be rejected, so adding that comment here would add confusion (That comment could be added in the controller function). An alternative implementation could be to regroup those parameters into a single parameter (ex: `filter`), which could be both a property or a lang, but then, unless creating a dedicated sanitization parameter, the sanitization would just validate that its a non-empty string, and property/lang validation would have to be done manually later.

server/controllers/entities/contributions.js
maxlath: fixed in f4717c7a

server/controllers/entities/contributions.js
maxlath: Alternative implementation attempted in f0c66ba and af12375, with a unique filter parameter accepting both properties and langs. This would have the advantage to be more easily extended later, for instance, if we were to include filters such as `merge` or `redirection` patches

jum-s (approved):

PR #588, opened 2021-12-17
Ignore activitypub delete activity

maxlath:
For instance, on 2021-12-17, the server encountered 1944 `could not fetch public key` errors, due to the fediverse sending `Delete` requests for actors that aren't followed from Inventaire (as following from Inventaire isn't implemented). By moving the `verifySignature` to the controllers that need it (currently only 1), we can first check if a signature is actually needed, and discard cases where the request can simply be ignored.


server/controllers/activitypub/inbox.js
```diff
@@ -20,8 +26,14 @@ const sanitization = {
   }
 }
 
-const controller = async params => {
-  return inboxActivityTypes[params.type](params)
+const controller = async (params, req) => {
+  const { type } = params
+  if (ignoredTypes.includes(type)) {
```
jum-s, line 31:
```suggestion
  if (!Object.keys(inboxActivityTypes).includes(type)) {
```

jum-s:
yes, `signed` is not a type of access (`public`, `authenticated`) as a request can be both (`signed` & `public`)

server/controllers/activitypub/inbox.js
maxlath: I prefer having an explicit list of ignored type, it means "we know that type but explicitly decide to not act on it", which is different from how an unknown type, already standardized or in a future extension, should be treated and reported

maxlath: Maybe that distinction isn't that important as both aren't supported anyway. Then we could just reject unknown/unsupported activity types at sanitization (before verifying the signature) with a 400: implementation proposed in 27d1bb8

jum-s (approved):

PR #595, opened 2022-01-11
remove faker

jum-s:


tests/api/fixtures/users.js
```diff
@@ -33,7 +33,7 @@ const API = module.exports = {
     return signup({
       email,
       username: API.createUsername(),
-      password: faker.internet.password()
+      password: randomBytes(8)
```
maxlath, line 36: `randomBytes` returns a buffer object
```suggestion
      password: randomBytes(8).toString('base64')
```

tests/unit/models/028-shelf.js
```diff
@@ -6,9 +6,9 @@ require('should')
 const someUserId = '1234567890a1234567890b1234567890'
 const { create, updateAttributes: update } = Shelf
 
-const faker = require('faker')
-const fakeName = faker.random.words(4)
-const fakeDesc = faker.random.words(15)
+const fakeText = require('./tests/api/fixtures/text')
```
maxlath, line 9:
```suggestion
const fakeText = require('tests/api/fixtures/text')
```

tests/api/fixtures/text.js
```diff
@@ -1,16 +1,81 @@
...
 
 const capitalize = word => word[0].toUpperCase() + word.slice(1).toLowerCase()
 
-const randomWords = (numberOfWords = 5) => faker.random.words(numberOfWords)
+const randomWords = (numberOfWords, suffix = '') => getSample(lorem, numberOfWords).join(' ').concat(suffix)
 
-module.exports = {
-  humanName: () => faker.fake('{{name.firstName}} {{name.lastName}}'),
+const getSample = (array, numberOfWords = 5) => {
+  // for loop looks ~3x faster than _.sample
```
maxlath, line 57: [non-blocking]
not such a hot path that a slight perf boost is worth the cost of maintaining our own helper function, imo

maxlath (changes requested):

tests/api/fixtures/users.js
jum-s: done in 3c173fedd

tests/unit/models/028-shelf.js
jum-s: done in 3c173fe

tests/api/fixtures/text.js
jum-s: done in 3c173fe

PR #600, opened 2022-01-26
refresh items snapshots after new entity operations

maxlath:
namely, `revert-edit` and `restore-version`, which were not taken into account, leading to obsolete snapshot data after those operations.

The other commits are just small cleanups while being around

### DIFF
--- a/tests/api/items/snapshot.test.js
+++ b/tests/api/items/snapshot.test.js
@@ -3,7 +3,7 @@ require('should')
 const { wait } = require('lib/promises')
 const { authReq } = require('../utils/utils')
 const { getById: getItem } = require('../utils/items')
-const { getByUris, merge, revertMerge, updateLabel, updateClaim } = require('../utils/entities')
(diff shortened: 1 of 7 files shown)


tests/api/items/snapshot.test.js
```diff
@@ -210,6 +210,34 @@ describe('items:snapshot', () => {
       reupdatedItem.snapshot['entity:authors'].should.equal(oldAuthors)
     })
 
+    it.only('should be updated when a related local entity is restored', async () => {
```
jum-s, line 213:
```suggestion
    it('should be updated when a related local entity is restored', async () => {
```

jum-s (changes requested):

tests/api/items/snapshot.test.js
maxlath: fixed in 1056590c

jum-s (approved):

PR #601, opened 2022-02-01
activitypub tweaks

maxlath:
attempting to reduce the amount of failed signature verifications

### DIFF
--- a/server/controllers/activitypub/lib/shared_key_pair.js
+++ b/server/controllers/activitypub/lib/shared_key_pair.js
@@ -7,12 +7,14 @@
 // cached keys will get refreshed at their next attempt
 // See https://socialhub.activitypub.rocks/t/caching-public-keys/688
 
-const { generateKeyPair } = require('lib/crypto').keyPair
+const { generateRsaKeyPair, sha1 } = require('lib/crypto')
 
 let sharedKeyPair
 
 const getSharedKeyPair = async () => {
-  sharedKeyPair = sharedKeyPair || await generateKeyPair()
+  if (sharedKeyPair) return sharedKeyPair
+  sharedKeyPair = await generateRsaKeyPair()
+  sharedKeyPair.publicKeyHash = sha1(sharedKeyPair.publicKey).slice(0, 10)
   return sharedKeyPair
 }
 
--- a/server/controllers/items/lib/items.js
+++ b/server/controllers/items/lib/items.js
@@ -101,8 +101,12 @@ const items_ = module.exports = {
     items = items.map(item => Item.create(userId, item))
     const res = await db.bulk(items)
     const itemsIds = _.map(res, 'id')
-    const { docs } = await db.fetch(itemsIds)
-    await emit('user:inventory:update', userId)
+    const shelvesIds = _.deepCompact(_.map(items, 'shelves'))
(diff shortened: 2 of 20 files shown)


server/controllers/activitypub/lib/shared_key_pair.js
```diff
@@ -7,12 +7,14 @@
...
-const { generateKeyPair } = require('lib/crypto').keyPair
+const { generateRsaKeyPair, sha1 } = require('lib/crypto')
 
 let sharedKeyPair
 
 const getSharedKeyPair = async () => {
-  sharedKeyPair = sharedKeyPair || await generateKeyPair()
+  if (sharedKeyPair) return sharedKeyPair
+  sharedKeyPair = await generateRsaKeyPair()
+  sharedKeyPair.publicKeyHash = sha1(sharedKeyPair.publicKey).slice(0, 10)
```
jum-s, line 17: why sha1 and not just a slice, or even the `publicKey` itself ?

maxlath: `sha1` is fast and this hashing is done only once per server process, while `publicKey` looks like `"-----BEGIN PUBLIC KEY-----\nMIICIjANBgkqhkiG9w0BAQEFAAOCAg8AMIICCg [...]` , and would need some transformation to be nice in a URL anyway

server/controllers/items/lib/items.js
```diff
@@ -101,8 +101,12 @@ const items_ = module.exports = {
     items = items.map(item => Item.create(userId, item))
     const res = await db.bulk(items)
     const itemsIds = _.map(res, 'id')
-    const { docs } = await db.fetch(itemsIds)
-    await emit('user:inventory:update', userId)
+    const shelvesIds = _.deepCompact(_.map(items, 'shelves'))
```
jum-s, line 104: i would make it `_.uniq`

maxlath: you also need the `flatten` (array of arrays) and the compact (some items might not have an array of shelves), which is what `deepCompact` does in addition to `uniq` (maybe `deepCompact` could be renamed to highlight that)

server/lib/sanitize/sanitize.js
```diff
@@ -46,6 +46,11 @@ const optionsNames = new Set([ 'nonJsonBody' ])
 const sanitizeParameter = (input, name, config, place, res) => {
   const parameter = getParameterFunctions(name, config.generic)
 
+  if (parameter.drop) {
+    delete input[name]
+    return
```
jum-s, line 51: a little test on drop in `tests/unit/libs/046-sanitize.js` would make me happy  [non-blocking]

maxlath: That's tested through the `ignore` generic parameter (see `it('should accept an ignored generic parameters'`). I started writing a more specific test, with something like:
```js
const configs = {
  name: {
    drop: true
  }
}
```
but that won't do, as that's the parameter controller-specific configuration and not the parameter global settings: we would need to have a parameter in `server/lib/sanitize/parameters.js` using `drop: true` to test that... which is what the `ignore`  generic does

jum-s (approved):

PR #603, opened 2022-02-17
add entities patch correction

maxlath:
Currently, when a user makes an entity edit (ex: updating a claim) and then corrects that edit, it creates 2 patches.

This PR implements a patch correction that would allow to reduce that to 1 patch, in the case where the entity is different at the end, or even 0 patch, in the case where the entity is brought back to its initial state.

Some case that might be more complex (ex: cases where an edit is done in a particular context, such as a merge or a batch) are ignored, to keep things simple for now.

This capacity to correct patches would be especially useful if we are to move to a UI that saves entity claims on the first click, rather than after a confirming click on the "save" button (as done in https://github.com/inventaire/inventaire-client/pull/311)


server/controllers/entities/lib/patches.js
```diff
@@ -63,12 +63,38 @@ module.exports = {
...
+        const beforeLastPatch = Patch.revert(currentDoc, previousPatch)
+        const aggregatedPatch = Patch.getDiff(beforeLastPatch, updatedDoc)
+        if (previousPatch.context == null && previousPatch.batch == null) {
+          if (aggregatedPatch.length === 0) {
+            await db.delete(previousPatch._id, previousPatch._rev)
+            return
+          } else if (aggregatedPatch.length < (previousPatch.patch.length + newPatch.patch.length)) {
+            previousPatch.patch = aggregatedPatch
+            return db.putAndReturn(previousPatch)
+          }
```
jum-s, line 80: this went from 2 lines to 5 `if` statements. Would it be possible to extract some logic to make it more readable ?

tests/api/entities/patches/patch_correction.test.js
```diff
@@ -0,0 +1,62 @@
...
+      await updateLabel({ uri: work.uri, lang: 'es', value: 'bar' })
+      const patches = await getHistory(work.uri)
+      patches.length.should.equal(2)
+      patches[1]._rev.should.startWith('2')
+      patches[1].patch.should.deepEqual([ { op: 'add', path: '/labels/es', value: 'bar' } ])
+    })
+  })
+
+  describe('delete', () => {
+    it('should delete a patch when the same user deletes the claim', async () => {
```
jum-s, line 46: what about adding the reverse test aka

`'should not delete a patch when another user deletes the claim`

maxlath: addressed in 3ecf869

server/controllers/entities/lib/patches.js
maxlath: I gave it a try in 640766a9, but couldn't really extract much, only try to make it less indigest

tests/api/entities/patches/patch_correction.test.js
```diff
@@ -0,0 +1,81 @@
...
+      patches.length.should.equal(1)
+    })
+
+    it('should not delete a patch when another user deletes the claim', async () => {
+      const [ userA, userB ] = await Promise.all([ getUserA(), getUserB() ])
+      const edition = await createEdition()
+      await updateClaim({ user: userA, uri: edition.uri, property: 'wdt:P1104', newValue: 5 })
+      await removeClaim({ user: userB, uri: edition.uri, property: 'wdt:P1104', value: 5 })
+      const patches = await getHistory(edition.uri)
+      console.log('🚀 ~ file: patch_correction.test.js ~ line 60 ~ it.only ~ patches', patches)
```
jum-s, line 60: useless line ?

PR #606, opened 2022-02-20
entities search: add pagination support

maxlath:
by accepting an offset parameter and returning total and continue values

### DIFF
--- a/server/controllers/search/search.js
+++ b/server/controllers/search/search.js
@@ -5,6 +5,7 @@ const { indexedTypes } = require('db/elasticsearch/indexes')
 const typeSearch = require('./lib/type_search')
 const Group = require('models/group')
 const { ControllerWrapper } = require('lib/controller_wrapper')
+const { addWarning } = require('lib/responses')
 
 const sanitization = {
   search: {
@@ -13,6 +14,7 @@ const sanitization = {
   lang: {},
   types: { allowlist: indexedTypes },
   limit: { default: 10, max: 100 },
+  offset: { default: 0, max: 500 },
   filter: {
     allowlist: [ 'wd', 'inv' ],
     optional: true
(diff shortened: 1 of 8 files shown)


server/controllers/search/search.js
```diff
@@ -32,21 +34,58 @@ const sanitization = {
...
 
+const entitiesSearch = async params => {
+  const { search, lang, limit, offset, claim } = params
+
+  if (!(_.isNonEmptyString(search) || _.isNonEmptyString(claim))) {
+    throw error_.newMissing('query', 'search or claim')
+  }
+
+  const { hits, total } = await typeSearch(params)
+  const continu = limit + offset
```
jum-s, line 81:
```suggestion
  const continue = limit + offset
```

tests/api/search/search_entities.test.js
```diff
@@ -219,6 +219,23 @@ describe('search:entities', () => {
         foundIds.should.not.containEql(humanB._id)
       })
     })
+
+    describe('offset', () => {
+      it('should offset results', async () => {
+        const types = 'works'
+        const input = 'memoire'
```
jum-s, line 226: this would mean some entities must have `memoire` as a label indexed in ES somewhere, which does not make this test isolated properly, i would be in favor of creating new entities for this test, or ensure entities are indexed somehow.

server/controllers/search/search.js
```diff
@@ -13,6 +14,7 @@ const sanitization = {
   lang: {},
   types: { allowlist: indexedTypes },
   limit: { default: 10, max: 100 },
+  offset: { default: 0, max: 500 },
```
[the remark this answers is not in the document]
maxlath: it's not the usual use-case to go that deep in the results, but once in a while, there might be a case where that could be useful to someone, maybe exploring the available data directly from the API: I would rather increase that arbitrary maximum than decrease it.

server/controllers/search/search.js
maxlath: can't do: `continue` is a reserved word in JS ^^'

tests/api/search/search_entities.test.js
maxlath: done in e8825f03

Issue #473, opened 2020-12-30
search: exact matches should work


maxlath:
ex: [wd:Q182847](https://inventaire.io/entity/wd:Q182847)

* [author: "Baudrillard"](https://inventaire.io/api/search?types=humans&search=Baudrillard&lang=fr&limit=50): wd:Q182847 comes 1st
* [author: "Jean Baudrillard"](https://inventaire.io/api/search?types=humans&search=Jean%20Baudrillard&lang=fr&limit=50): wd:Q182847 comes only 36th, the results being flooded by "Jean"s with [higher popularity scores](https://inventaire.io/api/entities?action=popularity&uris=wd:Q309240|wd:Q1343882|wd:Q574856|wd:Q3164657|wd:Q428160|wd:Q3170138|wd:Q3164666|wd:Q2708564|wd:Q123706|wd:Q3089290|wd:Q2705993|wd:Q2474670|wd:Q177411|wd:Q3165136|wd:Q1685793|wd:Q715206|wd:Q3167419|wd:Q123190|wd:Q3167174|wd:Q5829|wd:Q3168813|wd:Q3170128|wd:Q16267939|wd:Q6527|wd:Q707328|wd:Q1097651|wd:Q2329943|wd:Q179025|wd:Q773042|wd:Q27921811|wd:Q1685027|wd:Q1339|wd:Q233584|wd:Q3168741|wd:Q3167045|wd:Q182847), despite the exact label match

jum-s:
definitly. Exact match of the whole text search shall be boosted

maxlath:
adding the "critical" label after discussing with a data contributor who struggles to find the right authors, and sometimes then re-creates the author

jum-s:
I can put this afternoon on it i your fine with it ?

maxlath:
sure, I can't work on it now, so feel free to claim it

jum-s:
well exact match do work now

maxlath:
the local entities are easily not found despite an exact match, when their label share some words with Wikidata entities (despite the absence of popularity on those)

jum-s:
[opened a new issue](https://github.com/inventaire/inventaire/issues/485), as this one was labeled critical and was with an wikidata entity example

PR #608, opened 2022-02-22
entities search: prevent exact match to include non-exact matches due to edge n-gram tokens

maxlath:
This is done by indexing entities' labels and aliases a second time, but this time without applying the edge n-gram filter, to be able to perform exact search on them.

Example of a failed exact search (that should be fixed by this PR): [an exact search for "Norberto bob"](https://inventaire.io/api/search?types=humans&search=Norberto%20bob&lang=en&exact=true) should not find "Norberto Bobbio" (`wd:Q332693`)

While duplicating labels and aliases does have a cost in terms of index size, it would greatly improve features depending on exact search such as "merge suggestion" in the client or the resolver on the server.

This PR also addresses the single-word issue previously addressed in #517

### DIFF
--- a/server/controllers/search/lib/entities_query_builder.js
+++ b/server/controllers/search/lib/entities_query_builder.js
@@ -6,7 +6,7 @@ const { isPropertyUri, isWdEntityUri } = require('lib/boolean_validations')
 
 module.exports = params => {
   const { lang: userLang, search, limit: size, exact, claim, safe = false } = params
-  let { types, minScore = 1 } = params
+  let { types, minScore = 0.5 } = params
   types = getSingularTypes(types)
 
   const filters = [
@@ -19,7 +19,11 @@ module.exports = params => {
 
   if (!search) minScore = 0
 
-  const shoulds = matchEntities(search, userLang, exact, safe)
+  // In case a claim alone is searched
+  const shoulds = []
+  if (search != null) {
+    shoulds.push(matchEntities(search, userLang, exact, safe))
+  }
 
   return {
     query: {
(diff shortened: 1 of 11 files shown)


jum-s:
some production test strings can be found in issues #485 and #473

server/controllers/search/lib/entities_query_builder.js
```diff
@@ -48,68 +52,63 @@ module.exports = params => {
...
+        analyzer: 'standard_full',
+        type: 'best_fields',
       }
     }
-  ]
-
-  if (!exact) {
-    queries.push({
+  } else {
+    return {
```
jum-s, line 66: advantage of having the same multi_match for both exact and !exact was that exact must have a superior score to !exact (which is nice in every case (?)) 
drawback is exact performance, since 2 queries a fired

i would rather be for a little changes at a time and keep a same `multi_match` query for both exact and !exact matches

tests/api/search/search_entities.test.js
```diff
@@ -61,6 +64,13 @@ describe('search:entities', () => {
         })
       })
 
+      it('should not return results that do not contain the exact searched words', async () => {
```
jum-s, line 67: what about using the same wording as the reverse test right above this?
```suggestion
      it('should not return results that do not includes exact matches of each words', async () => {
```
[non-blocking]

jum-s, line 67: this test seems to be in contradiction with the comment line 53 

`// Ex: an exact search for "Charles Ben" will find "Charles Bent" as one of its generated tokens is "Ben"` 

?

jum-s: solved by b73e06d11

tests/api/fixtures/text.js
```diff
@@ -56,8 +56,19 @@ const randomWords = (numberOfWords, suffix = '') => _.sampleSize(lorem, numberOf
 
 const firstName = () => _.sampleSize(firstNames, 1)[0]
 
+const randomWord = () => randomWords(1)
+
+const distinctWords = () => {
```
jum-s, line 61: i would replace `distinctWords` by `randomWords` to only have `randomWords` which are distinct

server/controllers/search/lib/entities_query_builder.js
maxlath: done in 1372e549

tests/api/fixtures/text.js
maxlath: done in 3ca978e2

jum-s (approved):

PR #609, opened 2022-02-28
patches: do not anonymize patches from the requesting user

maxlath:
do not anonymize patches from the requesting user when requested by the history and contributions endpoints, so that the requesting user can see their contributions, even when their privacy settings do not let other user see them


tests/api/entities/contributions.test.js
```diff
@@ -136,6 +136,14 @@ describe('entities:contributions', () => {
       should(patches.find(isEntityPatch(workB)).user).not.be.ok()
       patches.find(isEntityPatch(workC)).user.should.equal(deanonymizedUser._id)
     })
+
+    it('should get non-anonimized contributions for the requesting user', async () => {
```
jum-s, line 140: inspired by `tests/api/entities/history.test.js `

```suggestion
    it('should not anonymize contributions when the author is the requesting user', async () => {
```

maxlath: done in 08f30038d

jum-s (approved):

Issue #610, opened 2022-03-03
Ĝ is a letter. (Or, usernames should be able to contain non-ASCII letters.)


JonathanReeve:
First: great job with Inventaire! I love it. Which is why I'd like to offer an opinion on how to make it better. 

I'd like to have my username be `ĝonatano` (my name). But when I try to enter that as a username, it says "username can only contain letters, figures (?), or `_`". But `ĝ` is very much a letter. Even as seen from an anglophone perspective, there's no denying it's a letter. 

![Ekrankopio de 2022-03-03 14-09-15](https://user-images.githubusercontent.com/1843676/156636630-45cecb2b-20aa-491f-b13f-691757244c8e.png)

This is an important issue, as I see it, since there are lots of names, words, username conventions, and so on that change their meanings entirely if accents are removed from their letters. In my language, `ĝ` is a qualitatively different letter from `g`: it's pronounced very differently. 

To understand why this is important, imagine your name is "John." Now imagine when you try to make your username "John123," a web app tells you "J" is not a letter. Maybe you can rewrite it as "ohn123" or "Gohn123," but you're either stuck mangling your name, or choosing something else more random, because it doesn't consider "J" to be a letter. 

FWIW, Wikidata allows non-ASCII letters in their usernames. I feel like if you have an international website, this is a necessity. 

[Here's what seems to be the code that does this.](https://github.com/inventaire/inventaire-client/blob/876f1d7f5c74f24246420cb113ea385bf113d6d0/app/modules/settings/components/profile.svelte#L60)

PR #611, opened 2022-03-05
allow unicode usernames

maxlath:
addressing https://github.com/inventaire/inventaire/issues/610

client PR: https://github.com/inventaire/inventaire-client/pull/315

### DIFF
--- a/server/models/user.js
+++ b/server/models/user.js
@@ -5,6 +5,7 @@ const error_ = require('lib/error/error')
 const randomString = require('lib/utils/random_string')
 const generateReadToken = randomString.bind(null, 32)
 const { truncateLatLng } = require('lib/geo')
+const { normalizeString } = require('lib/utils/base')
 
 const User = module.exports = {}
 
@@ -167,5 +168,6 @@ User.shouldBeAnonymized = user => {
 }
 
 User.formatters = {
-  position: truncateLatLng
+  username: normalizeString,
+  position: truncateLatLng,
 }
--- a/server/models/user.js
+++ b/server/models/user.js
@@ -17,6 +17,8 @@ User._create = (username, email, creationStrategy, language, password) => {
   assert_.strings([ username, email, creationStrategy ])
   if (language != null) { assert_.string(language) }
 
+  username = User.formatters.username(username)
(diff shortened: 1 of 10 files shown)


server/models/user.js
```diff
@@ -167,5 +168,6 @@ User.shouldBeAnonymized = user => {
 }
 
 User.formatters = {
-  position: truncateLatLng
+  username: normalizeString,
```
jum-s, line 171: as moved to the model, i would also move related api tests to integration tests, to not overtest api which is testing a standard JS function (`.normalize`)

maxlath: the goal of the tests I wrote was not to test `.normalize` itself, but that `.normalize` was indeed called

jum-s: I dont understand why there are model formaters, if its to repeat the formatting in the controllers sanitizer.
I would be in favor of only formatting username in the user model (i dont see any inconvenient so far).

tests/api/auth/signup.test.js
```diff
@@ -56,17 +52,42 @@ describe('auth:signup', () => {
...
-      email: `bla${username}@foo.bar`,
-      password: randomString(8)
+    await signup({ username })
+    .then(shouldNotBeCalled)
+    .catch(err => {
+      err.body.status_verbose.should.equal('an account is already in the process of being created with this username')
     })
+  })
+
+  it('should normalize unicode letters', async () => {
```
jum-s, line 66: as every string is formatted/normalized in the params sanitize, i think this test can be removed (which, i bet it was useful for development). 
Otherwise, every api tests involving a string parameter should make sure unicode is also normalized

maxlath: tests for other endpoints could be added, but it is much more critical for signup, due to username uniqueness constrain; I would thus not mind having only tests for endpoints where normalization is critical

jum-s: but this specific string is also normalized in the user model, so we could afford to not have one more api test if an integration test can do the job.
my point is that this test does not improve the test coverage, but well i guess this can be non blocking..

PR #525, opened 2021-05-05
NGI0 - Updating licensing aspects according REUSE 

lnceballosz:
Hello,

We have been working within the **NGI0** framework, helping projects with their licensing and copyright management. After a quick check on your repository,I would like to propose some updates regarding copyright and licensing information. [REUSE](https://reuse.software/) specifications intend to make licensing easier by setting one way to display all this legal information through comment headers on source files that can be human - and machine – readable.

The REUSE tool makes the process of applying licenses to files and compliance checking much easier. Check this screencast:

https://download.fsfe.org/NGI0/reuse.webm

Feel completely free to adopt these specifications by merging this pull request.

### REUSE Features:

- SPDX copyright and license comment headers for relevant files
- LICENSES directory with licenses used on the repository

 
### Files missing copyright and licensing information:

To serve as example, I added the SPDX headers with copyright and license information to the copyrightable files in the directory `server/* `with the exception of `server/db*` (dependencies) and `server/controllers* `(very large directory).

In large directories where including a comment header in each plain text file is impossible or undesirable, you may display copyright and licensing information in a [DEP5](https://reuse.software/spec/#dep5) file.

Please also double check if the personal information in the headers is correct and consistent. In the case of more copyright holders please fix the information in the headers.

I provided a more detailed license information to `README` file as well.

### Further REUSE compliance

In case you find REUSE useful, we offer a [wide range of tools](https://reuse.software/dev/) to help you to continuously check and display compliance with the REUSE guidelines.

Hope that helps and thank you very much for the amazing job!

PR #613, opened 2022-03-06
rename license into its own folder

jum-s:
according to #525 recommendations

in order to be able to close the PR #525


maxlath (changes requested):
we would also need to update the link in the README, maybe by reusing that line from https://github.com/inventaire/inventaire/commit/3056540427010aad49a3897661bc3a93d19f4e4e: Inventaire is an open-sourced project licensed under [AGPLv3](./LICENSES/AGPL-3.0-only.txt)

jum-s:
that was definitely included in the changes, but not committed...
done in 7582e342c

PR #618, opened 2022-03-31
make NodeJS core modules import explicit

maxlath:
by using the `node:` namespace
see https://nodejs.org/api/modules.html#modules_core_modules
or https://simonplend.com/make-your-imports-clearer-with-the-node-protocol/

This doesn't change anything other than making explicit what are core modules and what are third party modules.
Merging this PR would also mean that this is the new convention for core modules imports


maxlath:
that's more of a code gardening task, not worth the cognitive load of a PR

PR #625, opened 2022-05-06
make resolver nice

jum-s:
in order to let the server breath when resolving hundreds of uris with a long import (maybe prepmature optimization, but server have better things to do than resolving when under pressure)

### DIFF
--- a/server/controllers/tasks/collect_entities.js
+++ b/server/controllers/tasks/collect_entities.js
@@ -4,7 +4,7 @@ const db = require('db/couchdb/base')('entities')
 const { prefixifyInv } = require('controllers/entities/lib/prefix')
 const jobs_ = require('db/level/jobs')
 const checkEntity = require('./lib/check_entity')
-const { nice } = require('config').jobs['inv:deduplicate']
+const { nice } = require('config')
 const { waitForCPUsLoadToBeBelow } = require('lib/os')
 const batchLength = 1000
 
--- a/config/default.js
+++ b/config/default.js
@@ -203,9 +203,11 @@ const config = module.exports = {
   jobs: {
     'inv:deduplicate': {
       run: true,
-      nice: true,
     }
   },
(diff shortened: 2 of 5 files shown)


maxlath:
I think niceness should be reserved to non-interactive tasks, that is, tasks where no human is actively waiting for the result, in a similar way to the [`maxlag` recommendations in MediaWiki ](https://www.mediawiki.org/wiki/API:Etiquette#The_maxlag_parameter). There could be a parameter à la `maxlag` to determine that a request can wait: maybe something like `bot=true`?

Alternatively, if we are to use throttling for interactive tasks (such as the import which heavily uses the resolver), the load threshold should probably be higher: having a load threshold near 1 could make some sense, "you can do that resource intensive task as long as you are not sensibly slowing down other operations"?

server/controllers/tasks/collect_entities.js
```diff
@@ -58,7 +58,7 @@ const getFilteredUris = async (uris, refresh) => {
 
 const getUris = rows => _.map(rows, 'id').map(prefixifyInv)
 
-const deduplicateWorker = async (jobId, uri) => {
+const deduplicateWorker = async uri => {
```
maxlath, line 61: `deduplicateWorker` is called as `workerFn(jobId, payload)` (in `server/db/level/jobs.js`): this change breaks that call. Either revert change or change `workerFn(jobId, payload)` to `workerFn(payload)` (that shouldn't break anything else as `server/controllers/tasks/collect_entities.js` is the only consumer)

maxlath (changes requested):

server/controllers/tasks/collect_entities.js
jum-s: chose to left it like it is 
done in 9ff6f3bb

maxlath (approved):

PR #626, opened 2022-05-10
set single source of truth

jum-s:
adding security namespace in allowlisted parameter to not reject some
legit requests (ie. from user-agent 'Friendica')


tests/api/activitypub/followers_activity_delivery.test.js
```diff
@@ -29,13 +29,14 @@ describe('followers activity delivery', () => {
...
       createActivity['@context'].should.containEql('https://www.w3.org/ns/activitystreams')
+      createActivity['@context'].should.containEql('https://w3id.org/security/v1')
       createActivity.object.content.should.containEql(item._id)
       createActivity.to.should.deepEqual([ remoteUserId, 'Public' ])
     })
   })
 
   describe('entities followers', () => {
-    it('should post an activity to inbox', async () => {
+    xit('should post an activity to inbox', async () => {
```
maxlath, line 39: why disabling this test?

maxlath (commented):
I guess accepting that additional context for incoming requests should be fine, but I would rather not send it in our own outgoing requests, as we are not actually using that namespace, and who knows what other system might then raise similar errors due to that additional context(?)

jum-s:
Arent we using some vocab when signing request (ie. `crypto.createSign('rsa-sha256')`), for example `publicKeyPem` or `signature` defined `https://w3id.org/security/v1`?

FYI there is some info also at the bottom of [this](https://docs.joinmastodon.org/spec/security/) page, saying mastodon is using a draft namescpace while the LD spec says we should use `https://w3id.org/security/v2`.

tests/api/activitypub/followers_activity_delivery.test.js
jum-s: no idea how i ended up disabling it, but on `master` it was a red test. I had to fix it with 590caf345

maxlath:
as long as we are not using the `sec:` prefix in our signature, we don't need to declare that context, afaiu

jum-s:
done in 59bf4daf0

tests/api/activitypub/followers_activity_delivery.test.js
```diff
@@ -29,6 +29,7 @@ describe('followers activity delivery', () => {
       const { inbox } = await requests_.get(`${remoteHost}/inbox_inspection?username=${remoteUsername}`)
       const createActivity = inbox.find(a => a.type === 'Create')
       createActivity['@context'].should.containEql('https://www.w3.org/ns/activitystreams')
+      createActivity['@context'].should.containEql('https://w3id.org/security/v1')
```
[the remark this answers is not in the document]
jum-s: that was a commit amend mistake.... done in 89def0b7e

PR #629, opened 2022-05-23
Add support for actor url webfinger as resource

maxlath:
Currently, only resources in the form `acct:georges@inventaire.io` are accepted, rejecting resources in the form `https://inventaire.io/api/activitypub?action=actor&name=georges`

By accepting resources in the URL form, this PR would allow to answer with a 200 instead of a 400 to requests such as
```
GET /.well-known/webfinger?resource=https%3A%2F%2Finventaire.io%2Fapi%2Factivitypub%3Faction%3Dactor%26name%3Dgeorges 400
```
Such requests can be seen in the server logs, with Friendica in their user agent: I hope solving this would solve the issues reported lately (in mastodon DM) with Friendica integration

### DIFF
--- a/tests/api/activitypub/webfinger.test.js
+++ b/tests/api/activitypub/webfinger.test.js
@@ -128,15 +128,6 @@ describe('activitypub:webfinger', () => {
       const res2 = await publicReq('get', `${endpoint}${resourceAlias}`)
       res2.subject.should.equal(resource)
     })
-
-    it('should accept an actor URL as resource', async () => {
-      const username = createUsername()
-      await createUser({ username, fediversable: true })
-      const actorUrl = `${fullPublicHost}/api/activitypub?action=actor&name=${username}`
-      const res = await publicReq('get', `${endpoint}${encodeURIComponent(actorUrl)}`)
-      const { subject } = res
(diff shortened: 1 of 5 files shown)


tests/api/activitypub/webfinger.test.js
```diff
@@ -170,12 +187,22 @@ describe('activitypub:webfinger', () => {
...
-      res.should.be.an.Object()
       subject.should.equal(resource)
       const actorUrl = `${fullPublicHost}/api/activitypub?action=actor&name=${name}`
       decodeURIComponent(aliases[0]).should.equal(actorUrl)
       const firstLink = _.find(links, { rel: 'self' })
       decodeURIComponent(firstLink.href).should.equal(actorUrl)
     })
+
+    it('should accept an actor URL as resource', async () => {
+      const user = createUser({ fediversable: true })
```
jum-s, line 198: i dont think we need 3 duplicated tests (user shelf entity) for a format validation already defined in the unit test below. Since its not an core implementation. I would just leave one of those test in a specific `describe(''actor URL as resource'', () => {}`

maxlath: done in 7a6b25

PR #630, opened 2022-05-23
config: align URL part naming with WHATWG URL Standard

maxlath:
which is used by the global.URL constructor in NodeJS and window.URL in the browser

see https://developer.mozilla.org/en-US/docs/Web/API/URL/URL
and https://url.spec.whatwg.org/#url-class

Note: the branch is currently based on top of #629 to avoid conflicts. I considered including it in the #629 PR, but as that single commit touches a lot of files, that would have been more painful to review


config/default.js
```diff
@@ -9,25 +9,24 @@ const contactAddress = 'hello@inventaire.io'
...
-  fullHost: function () {
-    return `${this.protocol}://${this.host}:${this.port}`
+  getLocalOrigin: function () {
+    return `${this.protocol}://${this.hostname}:${this.port}`
   },
   publicProtocol: 'http',
-  publicHost: 'localhost',
-  fullPublicHost: function () {
-    return `${this.publicProtocol}://${this.publicHost}:${this.port}`
+  publicHostname: 'localhost',
```
jum-s, line 26: `publicHostname` & `hostname` have the same key value

maxlath: by default yes, but not in all environment: typically, our `config/local-production.js` file would contain:
```js
{
  publicHostname: 'inventaire.io'
}
```
while `hostname` would still use the default value : `localhost`. Should `hostname` be renamed `localHostname` or `privateHostname` to make the distinction clearer?

jum-s: i was more thinking of 
`publicHostname: config.hostname`
and environments who need a specific `publicHostname` key could define it in their config file

jum-s: ok

jum-s (approved):

PR #632, opened 2022-06-15
Items visibility

maxlath:
Like #624 but for items.

As items are a central concept, this PR touches to many things: shelves, transactions, feeds, activitypub, csv exports, email activity summaries, but those changes are mostly and fortunately superficial.

The heart of this PR is in:
- `server/db/couchdb/design_docs/items.js`
- `server/lib/visibility/allowed_visibility_keys.js`
- `server/controllers/items/lib/get_authorized_items.js`


server/controllers/items/lib/validate_item_async.js
```diff
@@ -2,6 +2,7 @@ const _ = require('builders/utils')
 const error_ = require('lib/error/error')
 const { validateVisibilityKeys } = require('lib/visibility/visibility')
 const getEntitiesByUris = require('controllers/entities/lib/get_entities_by_uris')
+const { flatMapUniq, mapUniq } = require('lib/utils/base')
```
jum-s, line 5: i thought the convention was to extract only if used by several consumers.
I would prefer to keep a expression such as `_.uniq(_.map(items, 'shelves').flat())` which is pretty straight forward to me

server/controllers/feeds/lib/generate_feed_from_feed_data.js
```diff
@@ -2,30 +2,25 @@ const _ = require('builders/utils')
...
 const snapshot_ = require('controllers/items/lib/snapshot/snapshot')
 const serializeFeed = require('./serialize_feed')
-const getItemsByAccessLevel = require('controllers/items/lib/get_by_access_level')
 const getAuthorizedItems = require('controllers/items/lib/get_authorized_items')
 const user_ = require('controllers/user/lib/user')
+const { filterPrivateAttributes } = require('controllers/items/lib/filter_private_attributes')
+const { paginate } = require('controllers/items/lib/queries_commons')
 
-module.exports = lang => async ({ accessLevel, reqUserId, feedOptions, users, shelves }) => {
+module.exports = lang => async ({ reqUserId, feedOptions, users, shelves, filter }) => {
```
jum-s, line 10: could we document what `filter` could mean in this context ?

server/controllers/items/by_entities.js
```diff
@@ -10,50 +10,12 @@ const sanitization = {
 }
 
 const controller = async params => {
-  return getEntitiesItems(params)
-  .then(addAssociatedData)
+  const { uris, reqUserId } = params
+  const foundItems = await items_.byEntities(uris)
+  const authorizedItems = await filterVisibleDocs(foundItems, reqUserId)
+  const page = paginate(authorizedItems, params)
```
jum-s, line 16: are we sure we can not do a more efficient request to the db by including a pagination within the view ? This looks costly, makes me sad a bit.

server/controllers/items/inventory_view.js
```diff
@@ -27,15 +27,14 @@ const validateUserOrGroup = params => {
...
 
 const getItems = async params => {
-  const { user, group, shelf, reqUserId, 'without-shelf': withoutShelf } = params
-  if (user) {
-    return getAuthorizedItems.byUser(user, reqUserId, { withoutShelf })
-  }
-  if (shelf) {
-    const shelfDoc = await shelves_.byId(shelf)
-    return getAuthorizedItems.byShelf(shelfDoc, reqUserId)
+  const { userId, groupId, shelfId, reqUserId, 'without-shelf': withoutShelf } = params
```
jum-s, line 30: Sanitization implicitly adding the suffix `Id` to `user` `group` and `shelf` is a bit confusing to me. "Where does that `Id` comes from ?" stopped me in the review, and i guess a newcomer could also block on this.

I would rely on JS convience to make it an explicit declaration: 

```suggestion
  const { user: userId, group: groupId, shelf: shelfId, reqUserId, 'without-shelf': withoutShelf } = params
```

server/controllers/items/lib/bulk_update.js
```diff
@@ -0,0 +1,45 @@
...
+  const formattedItems = currentItems.map(currentItem => Item.update(reqUserId, itemUpdateData, currentItem))
+  await validateValue({ attribute, value, reqUserId })
+  try {
+    const successfulUpdates = await items_.db.bulk(formattedItems)
+    await emit('user:inventory:update', reqUserId)
+    return previousUpdates.concat(successfulUpdates)
+  } catch (err) {
+    if (attempt > 10) throw err
+    const { body } = err.context
+    const [ failedUpdates, successfulUpdates ] = partition(body, hasError)
```
jum-s, line 21: [non-blocking] having to refer to line 4 to figure out that `partition` is a lodash function sucks. I preferred the previous inline notation `_.partition`. I believe we should be driven by our own readability, rather than what IDEs forces us to do.

server/controllers/items/lib/bulk_update.js
```diff
@@ -0,0 +1,45 @@
...
+    _.warn({ failedUpdates, successfulUpdates, attempt }, 'retrying bulk items update')
+    return bulkItemsUpdate({
+      reqUserId,
+      ids: failedUpdates.map(getId),
+      attribute,
```
jum-s, line 26: Documenting "why is `attribute` representing" could be helpful, since the below `visibility` and `shelves` refers to very different concept.

server/controllers/items/lib/validate_item_async.js
maxlath: several reasons push me toward preferring using the extract syntax by default:
- that's preparing the ground for the transition to the ES module syntax (already used on the client): those extracted variables will simply become named exports (ex: `export function flatMapUniq`), rather than key/values in a `export default` object
- there are cases where keeping a reference to an extracted variable rather than the object that contained it allows that object (and the other things it keeps a reference to) to be garbage collected, freeing memory. That would not be the case with `lib/utils/base`, as that object is anyway referenced all over the code base, but I think it's a good reason to have it as a default, allowing to not think to much about it
- my understanding is that there is a small performance benefit to do it like this: calling `_.uniq` rather than directly `uniq` means having to perform a lookup on the `_` object. The cost for that performance benefit being that we keep a reference to all those functions (`uniq`, `map`, etc) rather than just to their parent object (`_`)
- also, the auto-import in vscodium makes it easier to import like this

server/controllers/feeds/lib/generate_feed_from_feed_data.js
maxlath: Maybe `filter` is the wrong name here: what about renaming that key to `context`, which then lets the `paginate` function determine if a filter should be applied? Currently, the only use-case is when items from a group are requested, and that in that context, the requested private items are filtered out to not give the false impression that those private items appear in the group

server/controllers/items/inventory_view.js
maxlath: done in a8967657

server/controllers/items/lib/bulk_update.js
maxlath: See the other comment. I would be in favor of installing `@types/lodash` so that hovering the function gives you more info about what it is and its interface

server/controllers/items/lib/bulk_update.js
maxlath: both `visibility` and `shelves` arrays are attributes of the item doc, so that seemed self-explanatory enough to me, but if you can come up with something that addresses that issue, that would work for me too

server/controllers/feeds/lib/generate_feed_from_feed_data.js
jum-s: it does filter though, renaming is not necessary to me, a simple comment similar to the one in `get_by_users.test.js`
```suggestion
// filter known case: reject private items in group context to not give the false impression that those are visible to the other group members 
module.exports = lang => async ({ reqUserId, feedOptions, users, shelves, filter }) => {
```

server/controllers/items/lib/bulk_update.js
jum-s: relying on third party services, such as certain proprietary IDE features, is a dependency that makes it less stable than  plain old declaration in the beginning of the file.

server/controllers/feeds/lib/generate_feed_from_feed_data.js
jum-s: done in 61ebcbda9

jum-s (approved):

server/controllers/items/lib/bulk_update.js
maxlath: what proprietary feature?
- `@types/lodash` is MIT licensed
- vscodium and it's parent project are also MIT licensed
- typescript, which I assume is the language used for those type declarations, uses Apache License 2.0

jum-s: => proprietary "program" feature.
my point being, the more IDE agnostic the code is, the more stable. Its a dependency like another. 

Back on how to import this kind of external dependency functions, when i just see `partition` in the code, i would like to not have to refer to the dependency documentation and use a markup system like the current `_.foo()` to not have to think or hover or do whatever to find out what's this function anyway.

PS: note that i avoid trolling on who owns VSC & TS, since licence can change when they feel like it

maxlath:
I mistakenly merged the branch in `453-finer-shelves-privacy-settings` on which it was based, but remerged in master in 5dfd40d7

PR #636, opened 2022-07-25
resolver: allow to resolve editions without isbn or external id

maxlath:


tests/api/entities/resolver/non_strict_mode.test.js
```diff
@@ -4,11 +4,11 @@ const { randomLabel, generateIsbn13 } = require('tests/api/fixtures/entities')
 
 describe('entities:resolve:non-strict mode', () => {
   it('should ignore and report sanitization errors', async () => {
-    const entry = { edition: {} }
+    const entry = { edition: { isbn: '978000000000' } }
```
jum-s, line 7: 1. I still think documenting empty `edition: {}` is helpful.

2. Testing ` edition: { isbn: '978000000000' }` should return `invalid isbn`, since the requester does send an invalid isbn, which should throw even though some external id claims are found (?)

tests/api/entities/resolver/resolve.test.js
```diff
@@ -51,6 +51,56 @@ describe('entities:resolve', () => {
     entries[0].edition.uri.should.equal(`isbn:${isbn}`)
   })
 
+  it('should resolve an edition with no ISBN to an existing edition, if the title matches', async () => {
+    const { isbn, claims } = await createEditionWithIsbn()
+    const editionTitle = claims['wdt:P1476'][0]
```
jum-s, line 56: maybe adding a case insensitive feature

```suggestion
    const editionTitle = claims['wdt:P1476'][0].toUpperCase()
```

server/controllers/entities/lib/resolver/resolve_edition_from_works.js
```diff
@@ -0,0 +1,52 @@
...
+  const title = claims['wdt:P1476']?.[0]
+  return ASCIIFolder.foldMaintaining(normalizeString(title))
+}
+
+const editionSeedHasNoContradictingClaim = (editionSeed, edition) => {
+  for (const [ property, propertyClaims ] of Object.entries(editionSeed.claims)) {
+    if (!ignoredProperties.includes(property)) {
+      for (const value of propertyClaims) {
+        if (!edition.claims[property]?.includes(value)) {
+          return false
```
jum-s, line 41: getting in the case of contradicting edition kinda of suck right? Inventaire has some info, but the requester will not resolve its entry. I guess it would be nice to inform the requester, that there is something, so that the user can at least manually modify the edition in question, before resolving again?

PS: Same idea if multiple editions have been resolved (line 18 of this file), error information would be nice.

tests/api/entities/resolver/non_strict_mode.test.js
maxlath: 1. addressed in 8cdfa734f
2. It does return `invalid isbn` in the `res.errors`

tests/api/entities/resolver/resolve.test.js
maxlath: yes! done in 130cd7f9

server/controllers/entities/lib/resolver/resolve_edition_from_works.js
maxlath: I wouldn't say it sucks, its a best effort ^^ The worst case being that a new edition will be created, and maybe later merged in an existing edition. That seems acceptable to me, especially as the alternative would probably involves making the importer interface yet even more complex.

jum-s (approved):

PR #654, opened 2022-09-28
prevent the creation of several editions with the same isbn, when in autocreate mode

maxlath:
Making several calls to `GET /api/entities?action=by-uris` with `autocreate=true` can trigger the creation of duplicated edition entities. There seem to currently be more than 1600 in the production database. The reason for that is likely to be that the importer now seems to regularly trigger simultaneous requests for the same isbn-identified edition. Requests examples:

```
Sep 26 11:47:50 prod: GET /api/entities?action=by-uris&uris=isbn:9782738108074&refresh=false&autocreate=true 200 4465ms
Sep 26 11:47:50 prod: GET /api/entities?action=by-uris&uris=isbn:9782738108074&refresh=false&relatives=wdt:P629|wdt:P50&autocreate=true 200 664ms
```

which produces the following entities:
https://inventaire.io/entity/inv:76a0f40371636ef9184d3fdccd5778b5/history
https://inventaire.io/entity/inv:76a0f40371636ef9184d3fdccd578162/history

Investigating why simulatneous requests are sent by the client could make sense, but would then be less urgent if this problem is already solved on the server side.

The proposed implementation is to use a temporary memoization mechanism, so the first call to `getResolvedEntry` works as usual, and immediatly following calls get that same result promise. The memoization cache is then cleared once the first call returned, and after a short delay to let CouchDB views (which usually play the role of guard against duplicates) have the time to update.

Fortunately, the remediation to the current duplicates should be fairly simple: simply group the editions by isbns, and merge the duplicates.

### DIFF
--- a/server/data/dataseed/get_resolved_entry.js
+++ b/server/data/dataseed/get_resolved_entry.js
@@ -4,6 +4,7 @@ const { getByIsbns: getSeedsByIsbns } = require('./dataseed')
 const { enabled: dataseedEnabled } = require('config').dataseed
 const parseIsbn = require('server/lib/isbn/parse')
 const { resolvePublisher } = require('controllers/entities/lib/resolver/resolve_publisher')
+const temporarilyMemoize = require('lib/temporarily_memoize')
 
 const resolverParams = {
   create: true,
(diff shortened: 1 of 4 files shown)


jum-s (approved):

maxlath:
> Fortunately, the remediation to the current duplicates should be fairly simple: simply group the editions by isbns, and merge the duplicates.

Due to the peculiarities of the case, merging was actually not possible, and neither was soft deleting (=turning the docs into "removed_placeholders"), so I ended up hard deleting (=deleting the docs) the duplicated entities and their patches

PR #656, opened 2022-10-05
Implement context tailored listings

maxlath:
Currently, the main user private listings are displayed in every contexts, which can be confusing. This PR proposes to update the former `filter` parameter used for items into a `context` parameter, that takes a visibility key, and deduces what items or listings should not be returned for a given query, even if the requesting user has access rights.

### DIFF
--- a/tests/api/items/get_by_users.test.js
+++ b/tests/api/items/get_by_users.test.js
@@ -40,7 +40,7 @@ describe('items:get-by-users', () => {
       createItem(getUser(), { visibility: [ 'friends' ] }),
     ])
     const userId = groupsItem.owner
-    const { items } = await authReq('get', `${endpoint}&users=${userId}&filter=group`)
+    const { items } = await authReq('get', `${endpoint}&users=${userId}&context=group`)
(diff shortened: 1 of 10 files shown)


jum-s (changes requested):
Ive a test failing:

```
  1) listings:visibility
       groups
         should not return a groups-only listing to a friend:
```
looks like listing is returned on this branch

maxlath:
@jum-s that test passes for me, could you share the generated error stack trace?

tests/api/items/get_by_users.test.js
```diff
@@ -31,22 +32,24 @@ describe('items:get-by-users', () => {
     resItemsIds.should.containDeep(itemsIds)
   })
 
-  it('should not get private own user items in a group context', async () => {
-    // to avoid giving the false impression that those are visible by other members of the group
```
jum-s, line 35: maybe for newcomers, this comment could be of interest

jum-s: done in f126177b

jum-s (approved):

Issue #98, opened 2017-06-28
split update-claim into precise actions


maxlath:
I'm considering splitting `update-claim` into `add-claim`, `update-claim`, and `remove-claim`, as the [current implementation deduces the action from the existing parameters](https://github.com/inventaire/inventaire/blob/403889ef3085eabbd271de867b9fe35a1e7cac30/server/models/entity.coffee#L94-L109), which can thus have undesired consequences if a parameter name happen to be misspelled.

thewholelifetolearn:
I'm not sure that separating the actions is a good idea. That would mean that you should know before if you are adding or updating.  
After all, al actions are a kind of update. You update from *null* to `something`, from `something` to `something else` and from `something else` to *null*.  
Just submitting one value is fine. Don't need the old one.

maxlath:
I think having all in one endpoint is (slightly) more convenient but more error prone, especially for edition from the command line where it's easy to make a typo and thus trigger the wrong action, possibly even without noticing it: I would prefer having a more explicit syntax which would reject anything that doesn't look as expected.

You do need the old one for both *from `something` to `something else`* and *from `something else` to `null`.*

jum-s:
it's true that `update-claim` is a too-many-instructions function that you have to go all the way through if you ask why [this](https://github.com/inventaire/inventaire/blob/403889ef3085eabbd271de867b9fe35a1e7cac30/server/models/entity.coffee#L81)
No matter what, you create an entity & destroy it, I know beforehand those two actions to me
So, yeah max kill some logic here (only if necessary for future potential actions) it might ends with more readable code, to me

maxlath:
@jum-s true, even if I consider that the API clearness and ease of use have priority over internal code readabillity. Fortunately, in this case, I think both hint for the same direction.

PR #657, opened 2022-10-18
entities: update-label: allow to delete a label

maxlath:
This is currently not possible, which can be an issue in cases where a label was set in a language where there is no label for a given work. Example: [this work](https://inventaire.io/api/entities?action=by-uris&uris=inv:320d4c7c453886081a4da182e3d1e512) has a label in Catalan but no know label in Spanish, thus the attempt by a contributor to delete the Spanish label.

### DIFF
--- a/server/models/entity.js
+++ b/server/models/entity.js
@@ -49,7 +49,6 @@ const Entity = {
   setLabel: (doc, lang, value) => {
     assert_.object(doc)
     assert_.string(lang)
-    assert_.string(value)
 
     if (!wikimediaLanguageCodes.has(lang)) {
       throw error_.new('invalid lang', 400, { doc, lang, value })
@@ -57,13 +56,18 @@ const Entity = {
 
     Entity.preventRedirectionEdit(doc, 'setLabel')
 
-    value = _.superTrim(value)
+    if (value === null) {
+      deleteLabel(doc, lang)
+    } else {
+      assert_.string(value)
+      value = _.superTrim(value)
 
-    if (doc.labels[lang] === value) {
-      throw error_.new('already up-to-date', 400, { doc, lang, value })
+      if (doc.labels[lang] === value) {
+        throw error_.new('already up-to-date', 400, { doc, lang, value })
+      }
+
+      doc.labels[lang] = value
     }
-
-    doc.labels[lang] = value
 
     return doc
   },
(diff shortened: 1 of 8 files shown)


jum-s:
May i question why coming with a API change instead of a new action ?

Even though, this  updating to null = delete seems to exist elsewhere in the app, it explicitly are some explicit guard (like in [group model](https://github.com/inventaire/inventaire/blob/master/server/models/attributes/group.js#L20)).
A use case i see problematic is that forgetting to populate the new value in a request ends up in an unfortunate update which remove some data.
I kind of liked/got used to the explicit ways, like `updateClaim` with its `old-value`/`new-value` mechanism

Why not opening an explicit new action `deleteLabel` instead?

maxlath:
> forgetting to populate the new value in a request ends up in an unfortunate update which remove some data

no, because `null !== undefined`: not populating `value` would make it undefined, which is still rejected; it would have to be explicitly set to `null` to take effect

jum-s:
red test
```
  1) entities:update-labels
       should reject an up-to-date value:

      AssertionError: expected 429 to be 400
      + expected - actual

      -429
      +400
```

jum-s:
> it would have to be explicitly set to `null` to take effect

still more error prone than an explicit split . A topic that was interestingly discussed #98 some time ago.

jum-s:
i dont get why a [`canBeNull` option](https://github.com/inventaire/inventaire/blob/master/server/controllers/groups/update_settings.js#L9) is not necessary in the `updateLabel` controller. I guess it would document a bit that null is possible action

maxlath:
> i dont get why a [canBeNull option](https://github.com/inventaire/inventaire/blob/master/server/controllers/groups/update_settings.js#L9) is not necessary

because `optional=true`, but indeed `canBeNull` would be a better fit if we stay with that implementation

maxlath:
I tried to address the concern on the API in e55470631

server/models/entity.js
```diff
@@ -284,3 +288,15 @@ const setPossiblyEmptyPropertyArray = (doc, property, propertyArray) => {
     doc.claims[property] = propertyArray
   }
 }
+
+const deleteLabel = (doc, lang) => {
+  if (doc.labels[lang] == null) {
+    throw error_.new('can not delete a non-existant label', 400, { doc, lang })
```
jum-s, line 294:
```suggestion
    throw error_.new('can not delete a label which do not exist', 400, { doc, lang })
```

server/models/entity.js
```diff
@@ -284,3 +288,15 @@ const setPossiblyEmptyPropertyArray = (doc, property, propertyArray) => {
...
+  }
+
+  if (Object.keys(doc.labels).length === 1) {
+    throw error_.new('can not delete the last label', 400, { doc, lang })
```
jum-s, line 298:
```suggestion
    throw error_.new('can not delete the last remaining label', 400, { doc, lang })
```

jum-s:
> I tried to address the concern on the API in [e554706](https://github.com/inventaire/inventaire/commit/e55470631d2b767e67affb8db7b8e96c29b48ed4)

This commit doesnt belong to this PR, is that on purpose ?

maxlath:
@jum-s not sure what happened then, but after rebase, e554706 is now 6b2aa85 and is part of this PR

jum-s (approved):

PR #660, opened 2022-12-04
Add collection images special getter

jum-s:
Client publisher layout needs to illustrate a preview of collection covers:

Following the work and series way to do it

This is a make it work implementation.

A server delivering already crafted images object is nice, but at the cost of querying all editions... Maybe having the client request editions covers would make more sense to take advantage of caching the editions (?)

### DIFF
--- a/server/controllers/entities/lib/special_entity_images_getter.js
+++ b/server/controllers/entities/lib/special_entity_images_getter.js
@@ -24,9 +24,9 @@ module.exports = {
   },
 
   collection: async entity => {
-    entity.images = {}
+    const images = { claims: getEntityImagesFromClaims(entity) }
     return entities_.byClaim('wdt:P195', entity.uri, true, true)
-    .then(addEditionsImages(entity.images))
+    .then(addEditionsImages(images))
   }
 }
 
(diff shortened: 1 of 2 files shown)


server/controllers/entities/lib/special_entity_images_getter.js
```diff
@@ -21,6 +21,12 @@ module.exports = {
     const worksUris = _.map(parts, 'uri')
     const worksImages = await Promise.all(worksUris.map(getOneWorkImagePerLang))
     return worksImages.reduce(aggregateWorkImages, images)
+  },
+
+  collection: async entity => {
+    entity.images = {}
+    return entities_.byClaim('wdt:P195', entity.uri, true, true)
+    .then(addEditionsImages(entity.images))
```
maxlath, line 29:
```suggestion
    const images = { claims: getEntityImagesFromClaims(entity) }
    return entities_.byClaim('wdt:P195', entity.uri, true, true)
    .then(addEditionsImages(images))
```
`getEntityImagesFromClaims` allows to make images from Wikidata available on the  returned `images` object, which could provide good illustrations for collections old enough to be in the public domain, for instance.

maxlath (changes requested):

maxlath (approved):

PR #663, opened 2022-12-26
Be more tolerant on BNF ids format

maxlath:
Currently, if you enter a BNF id with its `cb` prefix (ex: `cb10317178p`), it is refused. This PR fixes that by adding a `format` function.


jum-s (approved):
Would be nice to inform the user with a field helper (text like `example: cd12345678 or 12345678`) in the editor. Because now no one can know that thanks to this PR cb prefix is accepted (ill open an issue on the client side).

maxlath:
I expect this branch to conflict with #664, so would wait for #664 merge to rebase and merge this one

PR #664, opened 2023-01-07
Convert codebase from CommonJS to ES modules

maxlath:
Several advantages to converting the codebase to [ES modules](https://nodejs.org/api/esm.html):
* It is now the standard, and everything will progressively push towards it
* It is possible to import CommonJS libs from an ESM codebase, but the reverse is not true (which, for instance, would prevent us from updating node-fetch to the latest version if we needed to)
* ESM allows to do named exports, which is very convenient
* Easier to parse by tools since dynamic imports are restricted, so we can expect IDE and linters to provide better DX and performance
* Consistent with the client
* Enables [top-level `await`](https://nodejs.org/api/esm.html#top-level-await)
* If we were to switch to TypeScript at some point, it would reduce the distance between source and generated code

Implementation details:
* I used [`cjs-to-es6`](https://www.npmjs.com/package/cjs-to-es6) to do the bulk of the conversion. This tool being unmaintained and not perfect, the output was unreliable/buggy in different places. Those shortcomings were then addressed in "cleanup" commits.
* Some heavy work went into modules that were exporting objects exposing functions to have those functions directly exported instead:
  * I used [this script](https://gist.github.com/maxlath/a8d1c03dfd3ccf3e1e0656ce35a4b62a) to reshape the export
  * and then did some renaming so that a function that used to be referred as `items_.byId` can now be automatically imported as `getItemById`. This also has the advantage to greatly increase the grepability. The only downside I encountered is that some test fixtures might already be using that name. If that were to be a problem, we could rely on some naming convention to have test fixtures function be named `fetchX` instead of `getX`, or something
* [`git-hooks`](https://www.npmjs.com/package/git-hooks) being in CommonJS, the hook scripts failed once package.json had `type=module`, so I replaced it with [`@vercel/git-hooks`](https://www.npmjs.com/package/@vercel/git-hooks)
* Some tests were failing due to bugs introduced in previous PRs. Those could have been solved on master, but would then likely have generated conflicts with this branch, so I included them here instead: 76aad8905 7df5fce97 262bff867 701c6d1ec 9506735e3


jum-s:
Could this PR have a different git user, a la `decaffeinate` user ([`67f037d`](https://github.com/inventaire/inventaire-client/commit/67f037d0cebd46558f7e442a5c077cc3509db5cf))?
This process was not without defaults, since accessing history was not straight forward, but it made git blaming easier. This is authoring is more of an issue for the server code since it doesnt look like it will be updated massively soon (unlike client Svelte rewrite).

jum-s:
to reduce the number line updated, what about not enforcing new comma on the last element of an iterable ? 
keep :
```
const sanitization = {
  name: {}
}
```
in order to not having this in the PR:
```
const sanitization = {
  name: {},
}
```
I guess a git diff + grep + xargs sed could do. Im fine doing it but waiting for answer before (you are also welcome to do it friend).

package.json
```diff
@@ -112,20 +114,17 @@
...
-    "models": "server/models",
-    "scripts": "scripts",
-    "server": "server",
-    "tests": "tests",
-    "root": "."
+  "imports": {
+    "#builders/*": "./server/builders/*.js",
+    "#controllers/*": "./server/controllers/*.js",
+    "#data/*": "./server/data/*.js",
+    "#db/*": "./server/db/*.js",
```
jum-s, line 121:
```suggestion
    "#db/*": "./server/db/*.js",
    "#fixtures/*": "./tests/api/fixtures/*.js",
```

.eslintrc.cjs
```diff
@@ -29,6 +29,7 @@ module.exports = {
     eqeqeq: [ 'error', 'smart' ],
     'implicit-arrow-linebreak': [ 'error', 'beside' ],
     'import/newline-after-import': 'error',
+    'import/no-useless-path-segments': [ 'error' ],
```
jum-s, line 32: For performance wise, `no-useless-path-segments` represents about 3.7% of total eslint time. Not a lot but significant enough to ask for a removal of this module that looks convenient for a one shot PR, but not significant enough, in my opinion, at every single linting (since it fixed 5 errors in years of development)

maxlath:
@jum-s 

> Could this PR have a different git user, a la decaffeinate user ([67f037d](https://github.com/inventaire/inventaire-client/commit/67f037d0cebd46558f7e442a5c077cc3509db5cf))? This process was not without defaults, since accessing history was not straight forward, but it made git blaming easier.

That commit is signed by `Author: decaffeinate <k@maxlath.eu>`, so it looks like the `decaffeinate` tool just temporarily overwrote the git config `user.name`. `cjs-to-es6` doesn't do that, but I think that's fine: the resulting commit (as well as several others from that PR) can be added to a `.git-blame-ignore-revs` file, see [*Ignoring bulk change commits with git blame*](https://moxio.com/blog/ignoring-bulk-change-commits-with-git-blame/). I started such a file when decaffeinating but apparently never committed it, I could do that now, and add the relevant commits from this PR once merged (so that we now the definitive commit hashes).

> This git authoring is more of an issue for the server code since it doesnt look like it will be updated massively soon (unlike client Svelte rewrite).

I didn't get that point.

maxlath:
@jum-s 

> to reduce the number line updated (hence preserving git history), what about not enforcing new comma on the last element of an iterable ?

If I recall correctly, having consistent commas was a requirement for [`convert_exports_default_object.js`](https://gist.github.com/maxlath/a8d1c03dfd3ccf3e1e0656ce35a4b62a). The corresponding hash could also be added to `.git-blame-ignore-revs` to work around the git history issue.

package.json
maxlath: done in a12d1ed32

.eslintrc.cjs
maxlath: done in d6da67f36

jum-s:
> can be added to a .git-blame-ignore-revs

yes, that would be great.

I would still be for a temporary author (ie. named `cjs-to-es6`) for these commits to not bias author statistics (ie. https://github.com/inventaire/inventaire/pulse).

jum-s:
> If I recall correctly, having consistent commas was a requirement for [`convert_exports_default_object.js`](https://gist.github.com/maxlath/a8d1c03dfd3ccf3e1e0656ce35a4b62a). The corresponding hash could also be added to `.git-blame-ignore-revs` to work around the git history issue.

Yeah, then it's fine like this.

jum-s:
Running `npm run indexation:wikidata:load-from-dump` on my machine throws:

```
Error [ERR_REQUIRE_ESM]: require() of ES Module inventaire/scripts/indexation/wikidata/format_dump_entity.js from node_modules/ndjson-apply/bin/ndjson-apply not supported.
```
Is that an expected behavior?

jum-s:
in local env, `package-lock.json` is not committed

maxlath:
@jum-s 
> Running npm run indexation:wikidata:load-from-dump on my machine throws ERR_REQUIRE_ESM

Fixed by a43f7b1

> in local env, package-lock.json is not committed

Fixed by 894526 (sorry for that, I'm experimenting with [pnpm](https://pnpm.io), which does not use `package-lock.json`; I have to think to run `npm i --package-lock-only` manually after modifying dependencies)

jum-s (approved):

Issue #377, opened 2020-04-04
index languages in Elasticsearch


maxlath:
today, auto-completion in language fields relies on [wikidata-lang](https://github.com/inventaire/wikidata-lang/) data (see client [language search providier](https://github.com/inventaire/inventaire-client/blob/27b9bb9/app/modules/entities/lib/search/language_search.coffee)). This has the advantage of limiting network request for what could be seen as a small set of data (there are way less languages than genres or works), but this means working with a limited set of data as we can't ship clients with languages labels in all languages, so we only use native and English labels. The consequence is that a French speaking person entering "Anglais" wouldn't be suggested any language.

The fix would be to index languages in ElasticSearch, just like we index other kinds of entities

maxlath:
Done in #699

Issue #198, opened 2018-06-20
increase the set of accepted languages in entities claims to include constructed languages


maxlath:
a user feedback:
>  I am trying to add Q10134 as edition language but it is not a supported entity

maxlath:
Done in #699

PR #670, opened 2023-02-14
Add an endpoint to find languages entities by code, term, or uri

maxlath:
Addressing #377 and #198

This PR does not fix https://github.com/inventaire/inventaire-client/issues/235, but lays foundations that could be used for that purpose

Client PR: https://github.com/inventaire/inventaire-client/pull/392

### DIFF
--- a/tests/api/entities/languages.test.js
+++ b/tests/api/entities/languages.test.js
@@ -33,4 +33,18 @@ describe('entities:languages:search', () => {
     const language = res.languages[0]
     language.uri.should.equal('wd:Q33491')
   })
+
+  it('should find a language by its wikidata entity id', async () => {
+    const res = await publicReq('get', `${endpoint}&search=Q10134`)
+    res.languages.should.be.an.Array()
+    const language = res.languages[0]
+    language.uri.should.equal('wd:Q10134')
+  })
+
+  it('should find a language by its wikidata entity uri', async () => {
+    const res = await publicReq('get', `${endpoint}&search=wd:Q10134`)
+    res.languages.should.be.an.Array()
+    const language = res.languages[0]
+    language.uri.should.equal('wd:Q10134')
+  })
 })
--- a/tests/api/entities/languages.test.js
+++ b/tests/api/entities/languages.test.js
@@ -21,17 +21,17 @@ describe('entities:languages:search', () => {
   })
 
   it('should find a language by its language code', async () => {
-    const res = await publicReq('get', `${endpoint}&search=hat&lang=zh`)
+    const res = await publicReq('get', `${endpoint}&search=nb`)
     res.languages.should.be.an.Array()
     const language = res.languages[0]
-    language.uri.should.equal('wd:Q33491')
+    language.uri.should.equal('wd:Q25167')
   })
 
   it('should ignore the code case', async () => {
-    const res = await publicReq('get', `${endpoint}&search=HAT&lang=zh`)
+    const res = await publicReq('get', `${endpoint}&search=Nb`)
     res.languages.should.be.an.Array()
     const language = res.languages[0]
-    language.uri.should.equal('wd:Q33491')
+    language.uri.should.equal('wd:Q25167')
   })
 
   it('should find a language by its wikidata entity id', async () => {
--- a/tests/api/entities/languages.test.js
+++ b/tests/api/entities/languages.test.js
@@ -1,0 +1,36 @@
+import 'should'
+import { publicReq } from '#tests/api/utils/utils'
+
+const endpoint = '/api/entities?action=languages'
+
+// The following tests require to have language entities indexed locally.
+// This can be done by running `npm run indexation:wikidata:load-languages-from-query`
+describe('entities:languages:search', () => {
+  it('should find a language by its label in the requested language', async () => {
+    const res = await publicReq('get', `${endpoint}&search=Korsische&lang=de`)
+    res.languages.should.be.an.Array()
+    const language = res.languages[0]
+    language.uri.should.equal('wd:Q33111')
+  })
+
+  it('should find a language by its label in another language', async () => {
+    const res = await publicReq('get', `${endpoint}&search=Korsische&lang=fr`)
+    res.languages.should.be.an.Array()
+    const language = res.languages[0]
+    language.uri.should.equal('wd:Q33111')
+  })
+
+  it('should find a language by its language code', async () => {
+    const res = await publicReq('get', `${endpoint}&search=hat&lang=zh`)
+    res.languages.should.be.an.Array()
+    const language = res.languages[0]
+    language.uri.should.equal('wd:Q33491')
+  })
+
+  it('should ignore the code case', async () => {
+    const res = await publicReq('get', `${endpoint}&search=HAT&lang=zh`)
+    res.languages.should.be.an.Array()
+    const language = res.languages[0]
+    language.uri.should.equal('wd:Q33491')
+  })
+})
--- a/docs/indexation.md
+++ b/docs/indexation.md
@@ -45,3 +45,8 @@ npm run indexation:load-from-couchdb entities
 ```sh
 npm run indexation:wikidata:load-from-dump
 ```
+
+### Wikidata languages entities
+```sh
+npm run indexation:wikidata:load-languages-from-query
+```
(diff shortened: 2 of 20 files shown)


jum-s:
What is the rational behind having a dedicated index instead of just having a type for language entities within `wikidata` index?  
I can see some advantages to process lang entities as "first class" wd entities: 

- It simplifies the code, as there would be no need for `hasNoType`
- This hypothetical `language` type would seem pretty distinct/independent from other possible types (ie. unlike `human` which also have organisations) which mean it would be quite future proof. 
- It doesnt split wikidata entities in two indexes, which was the reason behind having types on entities in the first place, no? Since the current implementation have a `wikidata` index with wikidata entities documents and `language` also with wikidata entities documents, which may be confusing.
- Lang entities could then be used as claim values (i can think of `P921` now, but future might have more occurrences of using those entities for other claims) without having to tweak editor autocomplete to search within two indexes, and it also simplifies editor autocomplete
- Creating an index for 9000 stable documents ("stable" as unlikely to be updated often) looks like an overkill.

Im limiting this comment to indexes, but it could be extended to the possibility of having 
`/api/search?types=languages&search=HAT&lang=en` which would make sense a priori for me (but maybe i dont see the whole picture), instead of a new endpoint that also does a ES `search` action on wd entities in the background.

tests/api/entities/languages.test.js
```diff
@@ -0,0 +1,50 @@
...
+
+  it('should find a language by its label in another language', async () => {
+    const res = await publicReq('get', `${endpoint}&search=Korsische&lang=fr`)
+    res.languages.should.be.an.Array()
+    const language = res.languages[0]
+    language.uri.should.equal('wd:Q33111')
+  })
+
+  it('should find a language by its language code', async () => {
+    const res = await publicReq('get', `${endpoint}&search=hat&lang=zh`)
```
jum-s, line 24: is the `lang` param necessary ? if not, removing it make the test clearer on what is necessary to pass the test

```suggestion
    const res = await publicReq('get', `${endpoint}&search=hat`)
```

tests/api/entities/languages.test.js
```diff
@@ -0,0 +1,50 @@
...
+    res.languages.should.be.an.Array()
+    const language = res.languages[0]
+    language.uri.should.equal('wd:Q33491')
+  })
+
+  it('should ignore the code case', async () => {
+    const res = await publicReq('get', `${endpoint}&search=HAT&lang=zh`)
```
jum-s, line 31: idem

tests/api/entities/languages.test.js
```diff
@@ -0,0 +1,50 @@
...
+describe('entities:languages:search', () => {
+  it('should find a language by its label in the requested language', async () => {
+    const res = await publicReq('get', `${endpoint}&search=Korsische&lang=de`)
+    res.languages.should.be.an.Array()
+    const language = res.languages[0]
+    language.uri.should.equal('wd:Q33111')
+  })
+
+  it('should find a language by its label in another language', async () => {
+    const res = await publicReq('get', `${endpoint}&search=Korsische&lang=fr`)
```
jum-s, line 17: just by looking at those test, ive difficulty to grasp what does the `lang` parameter brings to the request. I think it whould make since clearer to have a "negative" test which would document what not having `lang` param does.

maxlath:
> What is the rational behind having a dedicated index instead of just having a type for language entities within wikidata index?

The primary issue with handling languages as other indexed entities types is that the latter are typed using `getEntityType`, while I went on the idea to find languages by duck-typing ("any item that has a `wdt:P218|wdt:P219|wdt:P220|wdt:P1798|wdt:P221` claim is a language"), to avoid having to maintain yet another list of P31 values that should be considered languages ([for which there might currently be 122 candidates](https://query.wikidata.org/#SELECT%20%3Ftype%20%28COUNT%28%3Flanguage%29%20as%20%3Fcount%29%20%7B%0A%20%20%3Flanguage%20wdt%3AP218%7Cwdt%3AP219%7Cwdt%3AP220%7Cwdt%3AP1798%7Cwdt%3AP221%20%3Flanguage_id%20.%0A%20%20%3Flanguage%20wdt%3AP31%20%3Ftype%20.%0A%7D%0AGROUP%20BY%20%3Ftype%0AORDER%20BY%20DESC%28%3Fcount%29)).

> I can see some advantages to process lang entities as "first class" wd entities:
> * It simplifies the code, as there would be no need for hasNoType

It simplified that part, but complexifies the `/api/search` endpoint.

> * This hypothetical language type would seem pretty distinct/independent from other possible types (ie. unlike human which also have organisations) which mean it would be quite future proof.

Yes, that would be cleanly independent, but that would also mean we would need to handle the `type === 'language'` everywhere in the app where there is a switch on `type` (typically entity layouts). 

> * It doesnt split wikidata entities in two indexes, which was the reason behind having types on entities in the first place, no? Since the current implementation have a wikidata index with wikidata entities documents and language also with wikidata entities documents, it may be confusing.

The main reason to have entities share an index is to be able to search all those entities types in one query, but I don't see a use-case where we would search, for instance, for both works and languages. Languages won't be available in the global search bar. It can be a separated thing. (As could social search now, given how we segregate social vs entities in the search endpoint, but that's another story ><)

> * Lang entities could then be used as claim values (i can think of P921 now, but future might have more occurrences of using those entities for other claims)

It's already possible to use a language entity as P921 value: values for P921 are currently found by searching directly on Wikidata search API.

> * Creating an index for 9000 stable documents ("stable" as unlikely to be updated often) looks like an overkill.

That's one way to look at it. Another way is to see it as a nicely encapsulated domain, that won't have to suffer from changes in the way we handle other entities. (Such as when we will finally get back to [rewrite `getEntityType`](https://github.com/inventaire/inventaire/pull/294) ;))

> * It does not look like performance is an issue in case of storing langs entities in wikidata index (as [50Go index seem to be an accepted threshold](https://www.elastic.co/blog/how-many-shards-should-i-have-in-my-elasticsearch-cluster)).

Performance was indeed not a consideration here.

> Im limiting this comment to indexes, but it could be extended to the possibility of having /api/search?types=languages&search=HAT&lang=en which would make sense a priori for me (but maybe i dont see the whole picture), instead of a new endpoint that also does a ES search action on wd entities in the background.

The query performed on languages is different from the one performed on other entities (see `languagesQueryBuilder`), with primary purpose to give the right language as first result when entering a language code (da => wd:Q9035, nb => wd:Q25167, etc), so we would need to branch out somewhere anyway. 

So to summarize, yes it seems possible to have languages indexed with other wikidata entities, and to re-use the global search endpoint, but with some downsides:
* we need to maintain a list of languages P31s
* the search controller gets more complex and somewhat less friendly: requesting `/api/search?types=works|languages` should return an error, as it relies on 2 different queries  (just like `/api/search?types=works|groups`)
* future work on entities typing will need to care for languages too, contributing to make that task even more daunting

tests/api/entities/languages.test.js
maxlath: removed in c51fc9f63

tests/api/entities/languages.test.js
maxlath: removed in c51fc9f63

tests/api/entities/languages.test.js
maxlath: The `lang` parameter should just boost some results if they happen to match on the requested language. The negative test would get us deep in the questions of Elasticsearch scoring; I would rather not go there, as that seems to be a receipt for a fragile test.

maxlath:
TODO:
- [ ] add language code property: [IETF language tag (P305)](https://www.wikidata.org/wiki/Property:P305)

tests/api/entities/languages.test.js
jum-s: Would something like:

```js
it('should find english as default language', async () => {
  const res = await publicReq('get', `${endpoint}&search=Italian`)
  const language = res.languages[0]
  language.uri.should.equal('wd:Q652')
})
```

be subjected to results that would be inconsistent/not future proof ?

This test would not document the purpose of this `lang` parameter (thus digging in ES scoring), but at least we can know that it's optional (?)

Also we could include your description:

> The `lang` parameter should just boost some results if they happen to match on the requested language. 

in a top comment of the test file.

maxlath:
Closing in favor of #699

PR #673, opened 2023-03-27
get_bnf_entry_from_isbn: add edition contributors

maxlath:
Namely: translators, preface and postface authors

### DIFF
--- a/server/controllers/entities/lib/resolver/resolve_external_ids.js
+++ b/server/controllers/entities/lib/resolver/resolve_external_ids.js
@@ -6,7 +6,7 @@ import { forceArray } from '#lib/utils/base'
 import getInvEntityCanonicalUri from '../get_inv_entity_canonical_uri.js'
 import properties from '../properties/properties_values_constraints.js'
 
-export default async (claims, resolveOnWikidata = true) => {
+export async function resolveExternalIds (claims, resolveOnWikidata = true) {
   const externalIds = []
 
   for (const prop in claims) {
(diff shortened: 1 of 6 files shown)


server/controllers/entities/lib/resolver/resolve_external_ids.js
```diff
@@ -6,7 +6,7 @@ import { forceArray } from '#lib/utils/base'
 import getInvEntityCanonicalUri from '../get_inv_entity_canonical_uri.js'
 import properties from '../properties/properties_values_constraints.js'
 
-export default async (claims, resolveOnWikidata = true) => {
+export async function resolveExternalIds (claims, resolveOnWikidata = true) {
```
jum-s, line 9: ah thx for that, much more greppable with named function

server/data/bnf/helpers.js
```diff
@@ -0,0 +1,26 @@
+import { compact, uniq } from 'lodash-es'
+import { resolveExternalIds } from '#controllers/entities/lib/resolver/resolve_external_ids'
+import { isWdEntityUri } from '#lib/boolean_validations'
+
+export async function setEditionContributors (edition, property, urls) {
+  const uris = await resolveContributors(urls)
+  if (uris.length > 0) {
+    edition.claims[property] = uris
```
jum-s, line 8: There is a common pattern of seeds sometimes assigning translators as P50. But  this will not conflict with the possibility of already having those authors in the P50 claim, right? aka BNF based entity creation is not based on the seed, right?

jum-s:
wanted to rebase and merge, but tests need a functionnal BNF API which is not the case at the moment

jum-s (approved):

PR #676, opened 2023-04-16
Propagate listing elements redirections

jum-s:
A command line to check if there are redirections missing in prod

`curljson "LH:5984/elements-tests/_all_docs?include_docs=true" | jq -r ".rows[].doc.uri|." | xargs -I'@@' curl -H 'Content-Type:application/json' -H 'Accept: application/json' "LH:3009/api/entities?action=by-uris&uris={@@}" | jq '.redirects|select(length > 0)'`


maxlath:
Propagation in prod done :heavy_check_mark:

PR #686, opened 2023-05-25
/api/entities?action=by-uris: attach entities popularities when requested

maxlath:
This would allow to simplify code such as `getAndAssignPopularity` in https://github.com/inventaire/inventaire-client/pull/384 by getting popularities straight from the call to `getEntitiesAttributesByUris`:

ex: `GET /api/entities?action=by-uris&attributes=type|labels|popularity&uris=wd:Q535`

### DIFF
--- a/server/controllers/entities/by_uris_get.js
+++ b/server/controllers/entities/by_uris_get.js
@@ -1,3 +1,4 @@
+import { addEntitiesPopularities } from '#controllers/entities/lib/popularity'
 import addRelatives from './lib/add_relatives.js'
 import getEntitiesByUris from './lib/get_entities_by_uris.js'
 import { pickAttributes, pickLanguages } from './lib/pick_attributes.js'
@@ -12,6 +13,7 @@ const sanitization = {
       'claims',
       'sitelinks',
       'image',
+      'popularity',
     ],
     optional: true,
   },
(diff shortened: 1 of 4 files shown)


jum-s:
The client dev in me is quite happy about the PR: it does make the code simpler, but, the back dev is crying a little bit, it feels like hiding a lot of dust under the carpet.

The question on the long run is: if we merge this now and wait a few years, will it be fluid for client devs to ask for many popularity requests at once, without worrying about overcrowding the server?

maxlath:
> will it be fluid for client devs to ask for many popularity requests at once, without worrying about overcrowding the server

Unless the `GET /api/entities?action=by-uris` request includes `refresh=true`, a request for many non-cached popularity will simply populate the `popularityJobQueue`, which `waitForCPUsLoadToBeBelow({ threshold: 0.4 })` before running a job, so that should not be able to overwhelm the server

jum-s:
In the future, nothing will forbids a client to ask thousands of popularities with refresh true?

jum-s:
> @jum-s Unless the `GET /api/entities?action=by-uris` request includes `refresh=true`, a request for many non-cached popularity will simply populate the `popularityJobQueue`, which `waitForCPUsLoadToBeBelow({ threshold: 0.4 })` before running a job, so that should not be able to overwhelm the server

It looks acceptable to add attributes `popularity`, but it would be nice to keep the possibility from the client to keep a way to only request `popularity` key (via `API.entities.popularity` path), in order to refresh the popularity without refreshing the whole entity.

maxlath:
> nothing will forbids a client to ask thousands of popularities with refresh true

Yes. That's also already the case with `/api/entities?action=popularity&uris={uris}&refresh=true`. I think this is not as big of a problem as that might have been in the past, as there is now a job queue that takes care of smoothening the resource use. We could look into rate limiting if there was resource abuse, but that seems overengineered for now. Alternatively, if that was to become a problem, we could ignore the `GET /api/entities?action=by-uris&attributes=type|labels|popularity&uris=${uri}&refresh=true` refresh flag for the `popularity` attribute

PR #698, opened 2023-09-18
share entities properties metadata with client

maxlath:
in order to ease the client work to stay up-to-date with available properties, their datatypes, and constraints (ex: single or multiple values allowed, entity type of allowed values)

Addressing #112 
~Based on #696 as both PR modify the same files, but could be made independent if needed~
Corresponding client PR https://github.com/inventaire/inventaire-client/pull/437 , where this PR allows to get rid of some "keep in sync with server" comments

PR #699, opened 2023-09-18
add languages to searchable entities

maxlath:
This PR is a reboot of #670 following the implementation proposed in [this comment](https://github.com/inventaire/inventaire/pull/670#issuecomment-1432790518). It does simplify things, but at the cost of letting some languages out (namely those that don't have a P31 matching the language alias list  [added in `d4271c3`](https://github.com/inventaire/inventaire/commit/d4271c342#diff-df75609b219c6dbc6401216b0f137269b1e2bb52719c5778365cc2ff0a4c4ad6))

Addressing #377 and #198

Based on #698 as that where https://github.com/inventaire/inventaire/commit/d4271c342 adds the languages P31 alias list.

This PR still doesn't fix https://github.com/inventaire/inventaire-client/issues/235, but still lays foundations that could be used for that purpose.


server/controllers/entities/lib/build_popularity_by_uri.js
```diff
@@ -77,12 +77,23 @@ const getEntitiesPopularityTotal = uris => {
   .then(results => _.sum(results) + results.length)
 }
 
+const getTypeAgnosticEntityScore = async uri => {
```
jum-s, line 80: That can be useful indeed (and fast). 
A naming proposition since those entities have a P31, and the client does categorize those with a type derived from this wikidata's P31. While what differs from other popularity builders is the fact that it doesnt query any entity graph, isn't it ? What about `getSimpleEntityScore` or `getChildAgnosticEntityScore`

[non-blocking]

tests/api/search/search_languages_entities.test.js
```diff
@@ -0,0 +1,42 @@
...
+import { search, waitForIndexation } from '#tests/api/utils/search'
+import 'should'
+
+const wikidataUris = [ 'wd:Q10134', 'wd:Q19852850', 'wd:Q33111' ]
+
+describe('entities:languages:search', () => {
+  before(async () => {
+    // Ensure wikidata uris are indexed in the current format
+    await getByUris(wikidataUris, null, true)
+    await Promise.all(wikidataUris.map(uri => waitForIndexation('wikidata', uri.split(':')[1])))
```
jum-s, line 11: What about reusing `findOrIndexEntities`?

```suggestion
    await findOrIndexEntities(wikidataUris)
```

server/controllers/search/lib/entities_query_builder.js
```diff
@@ -157,3 +161,14 @@ const validatePropertyAndValue = condition => {
...
   }
 }
+
+function matchLanguageCode (search) {
+  search = search.toLowerCase()
+  const claimTerms = languagesCodesProperties.map(property => `${property}=${search}`)
+  return {
+    terms: {
+      claim: claimTerms,
+      boost: 100,
```
jum-s, line 171: Isnt this `boost` an overkill ? Since, for the moment, the client is not searching with both a type `languages` and another type. Even if it does one day, i guess one would also like the exact match of an [entity name with a language string in it](inventaire.io/entity/wd:Q271931).

server/controllers/entities/lib/build_popularity_by_uri.js
maxlath: renamed to `getSimpleEntityScore` in 10a527607

tests/api/search/search_languages_entities.test.js
maxlath: Indeed! Done in 85e699396

jum-s (approved):
To be merged after #698

Issue #682, opened 2023-05-08
add types and automatic API documentation


maxlath:
We started discussing stronger typing back when we transitioned away from CoffeeScript #72, and since then the TypeScript boat got a lot of wind in its sails. It now seems clear that the way to go is either TypeScript or a TypeScript-based annotation system, typically [JSDoc](https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html), possibly until [TS becomes standardized within JS](https://github.com/tc39/proposal-type-annotations). Whatever the progress of that proposal, it can't hurt to start adding typing as JSDoc, which could later either be converted to TS or to TS-ified JS.

#### pros
* No need to add a compilation step. The server executable remains `node`, rather than `ts-node` or [`swc-node`](https://github.com/swc-project/swc-node). In the future, moving to a runtime fully integrating TS, such as [Deno](https://deno.com) or [Bun](https://bun.sh/), could be an option, but the transition cost could be high, and there is no clear winner yet.
* No transition cost: we can start today, change our mind in either direction tomorrow, at small cost
#### cons
* Requires to learn the "typing in JSDoc" particularities, additionally to learning TS
* Less practical and less well integrated than TS(?)

#### pros
* More complete(?)
* Simpler syntax than when having to fiddle with comments
#### cons
* transpilation:
   * Slower dev (re)start due to re-transpilation (type check can be disabled by using `ts-node --swc`)
   * prod shouldn't be affected too much, if files are compiled ahead of restarting, and the server then starts from the built .js files. But then we loose/complexify stack traces, or have to fiddle with source maps?

### Runtime type checking and automatic API documentation
The reason I want to link the subject of API documentation with types is that I do think they should be considered together, in particular having in mind what a tool such as [`zod`](https://zod.dev/) (thanks @douginamug for bringing my attention to it), or one of its alternatives ([arktype](https://github.com/arktypeio/arktype), [typebox](https://github.com/sinclairzx81/typebox)), could offer:
* We could replace our current models validation and API input sanitization by zod schemas, which would handle the runtime validation, and from which types could be inferred (which [should also work in JSDoc](https://blog.jim-nielsen.com/2023/types-in-jsdoc-with-zod/)). See also [`zod-validation-error`](https://github.com/causaly/zod-validation-error)
[...]

PR #706, opened 2023-12-03
add Typescript

maxlath:
Proposes to start addressing #682 by adding Typescript tools. It seems possible to keep the cost of transpilation to an acceptable level:
- dev: the watcher is sensibly slower than with vanilla JS (between 3 and 5s in my mesurement)
- prod: pre-building files allows having zero restart time cost

I would have prefered to rename all files to `.ts` to avoid to deal with renaming issues during the progressive adoption of types, but then the type check errors would crash the build (despite `noEmitOnError=false` in `tsconfig.json#compilerOptions`). The build works when files stay in JS, as long as the build is done with `checkJs=false`, so I would suggest we keep it like this until we can get to the point where no type check error blocks the build.

~~PR based on #705~~


tests/api/items/snapshot.test.js
```diff
@@ -25,7 +25,7 @@ import { getItem } from '../utils/items.js'
 import { authReq, getUserB } from '../utils/utils.js'
 
 describe('items:snapshot', () => {
-  it("should snapshot the item's work series names", async () => {
+  it.only("should snapshot the item's work series names", async () => {
```
jum-s, line 28: ?

maxlath: That's a mistake. Fixed in 70b945f64

server/controllers/entities/lib/get_entity_by_uri.ts
```diff
@@ -0,0 +1,15 @@
+import { getEntitiesByUris } from './get_entities_by_uris.js'
+
+interface GetEntityByUriArgs {
+  uri: string
```
jum-s, line 4: it looks like there is an attempt in 044fc55 to change file extension `.js`->`.ts`
despite PR comment
> I would have prefered to rename all files to .ts ...

The new ts file `get_entities_by_uris.ts` has not been typed (ie. `response.entities = keyBy(response.entities, 'uri')`), thus ending up crashing the build.

Shall `get_entities_by_uris` be renamed with `.js` extension ?

PR #726, opened 2024-02-05
get Openlibrary summaries by ISBN

maxlath:
In the current implementation, OpenLibrary summaries are only fetched when there is a know OpenLibrary identifier, but summaries for editions could also be fetched by ISBN

### DIFF
--- a/server/controllers/data/lib/summaries/openlibrary.js
+++ b/server/controllers/data/lib/summaries/openlibrary.js
@@ -1,9 +1,10 @@
 import { cache_ } from '#lib/cache'
+import { normalizeIsbn } from '#lib/isbn/isbn'
 import { requests_ } from '#lib/requests'
 
 const timeout = 10 * 1000
 
-export default async ({ id, refresh }) => {
+export async function getOpenLibrarySummaryByOpenLibraryId ({ id, refresh }) {
   const lastLetter = id.slice(-1)[0]
   const section = openLibrarySectionByLetter[lastLetter]
   const link = `https://openlibrary.org/${section}/${id}`
@@ -34,4 +35,29 @@ const openLibrarySectionByLetter = {
   A: 'authors',
   W: 'works',
   M: 'books',
+}
+
+export async function getOpenLibrarySummaryByIsbn ({ id, refresh }) {
+  const isbn = normalizeIsbn(id)
+  const link = `https://openlibrary.org/isbn/${isbn}`
+  const url = `${link}.json`
+  const property = 'wdt:P212'
+  const text = await cache_.get({
+    key: `summary:${property}:ol:${isbn}`,
+    refresh,
+    fn: async () => {
+      const { description: text } = await requests_.get(url, { timeout })
+      if (!text) return
+      if (text.value) return text.value
+      else if (typeof text === 'string') return text
+    },
+  })
+  if (text) {
+    return {
+      text,
+      name: 'OpenLibrary',
+      link,
+      lang: 'en',
+    }
+  }
 }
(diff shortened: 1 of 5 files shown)


tests/api/data/summaries.test.js
```diff
@@ -37,6 +38,21 @@ describe('summaries', () => {
       summaryData.lang.should.equal('en')
     })
 
+    it('should detect summary language', async () => {
```
jum-s, line 41: some red tests

```
should return summaries:
     TypeError: Cannot read properties of undefined (reading 'key')
      at Context.<anonymous> (file:///home/vinny/inventaire/docker/inventaire/tests/api/data/summaries.test.js:32:19)
      at processTicksAndRejections (node:internal/process/task_queues:96:5)
```

maxlath: @jum-s this might be caused by the change cache data format, can you try to run the tests in non-quick mode (or just run `./scripts/tests/api/delete_tests_databases.sh`)  to reset the leveldb before running the tests?

jum-s: yep, thanks

jum-s: Most tests passes now after resetting level db correctly (which script `delete-databases` didn't do properly (see fix cabf38154)).
But there is still:

```
  12 passing (7s)
  1 failing

  1) summaries
       openlibrary
         should ignore unknown ISBNs:
     request error: invalid uri: expected uri, got  (string)
```

maxlath: Two problems there:
- `existsOrCreate` was silencing the actual error, fixed by 924c46bc2
- This test uses an ISBN considered valid only in the most recent versions of `isbn3`, which was updated in 424824f59 but we don't have a hook to make other developers rerun `npm install` so you didn't get the update: just running `npm install` should fix the issue

jum-s: What about having a warning with a git hook: 

`.git/hooks/post-checkout`

> invoked when a git checkout is run after having updated the worktree


```sh
CHANGED=`git diff $1 $2 --stat package-lock.json`
if [[ -n $CHANGED ]]
then
    echo "△  Warning: package.json has changed."
    echo "△  You may want to run 'npm install'."
fi
```

maxlath:
@jum-s good to merge?

server/controllers/data/lib/summaries/openlibrary.js
```diff
@@ -1,31 +1,68 @@
...
+    cacheKey = `summary:${key}:${id}`
+    claim = { id, property }
+  } else if (isbn13h) {
+    const isbn13 = normalizeIsbn(isbn13h)
+    link = `https://openlibrary.org/isbn/${isbn13}`
+    url = `${link}.json`
+    property = 'wdt:P212'
+    key = `${property}:openlibrary`
+    cacheKey = `summary:${key}:${isbn13}`
+    claim = { id: isbn13h, property }
```
jum-s, line 30: Could it make sense to add a `ttl` ? In case the ISBN cache key is not in use anymore. ie a fresh OL id has recently been added to the entity.

server/controllers/data/lib/summaries/openlibrary.js
```diff
@@ -1,31 +1,68 @@
...
     fn: async () => {
-      const { bio, description } = await requests_.get(url, { timeout })
-      const text = bio || description
-      if (!text) return
-      if (text.value) return text.value
-      else if (typeof text === 'string') return text
+      try {
+        const { bio, description, languages } = await requests_.get(url, { timeout })
+        const text = bio || description
+        if (!text) return
```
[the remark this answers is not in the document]
maxlath: OpenLibrary is kind enough to accept ISBN-13 in place of ISBN-10, see for your first example https://openlibrary.org/isbn/9780517588475. But your examples raises an other issue: the edition itself doesn't have a description, but its work does, and it would be nice to get that instead of nothing.

server/controllers/data/lib/summaries/openlibrary.js
maxlath: By default `cache_.get` will give a ttl equal to `CONFIG.leveldb.defaultCacheTtl`

PR #734, opened 2024-02-22
get_entities_by_isbns: refresh seeds when requested

maxlath:
Currently, external databases (aka authorities) are checked when there is no entity in the local database for a given ISBN. The proposed change would be to re-check those external databases for existing edition entities when requested, that is when `autocreate=true&refresh=true`.

Writing a test is unfortunately not so straight forward, as you would need to create an edition/work/author with the right labels, so that the resolver can confidently update those entities. But as I expect it to become a popular feature from data admins, I'm sure we will notice if this breaks somehow ;)

Client PR https://github.com/inventaire/inventaire-client/pull/482


jum-s (approved):
`getResolvedEntry` has a confusing naming, since it doesn't only get but also set some data. It's out of scope of this PR (hence only this comment), but I would be for having a stricter policy: `get*` would be reserved for functions that do not write, only fetch stuff.

maxlath:
rename attempted in 40b4b78

Issue #724, opened 2024-02-01
Display remaining tasks overview


jum-s:
In today's codebase: 
- there is only one task type (`deduplicate`) stored in db
- `suspectUri` and `suggestionUri` pair is unique
- there are two kinds of `deduplicate` tasks, based on their entity type (`human` and `work`)
- all tasks with `human` entities are generated automatically, which currently creates a lot of tasks 
- all tasks with `work` entities are based on user feedback ; those have a `reporter: userId` 

The objective is to have users access a dashboard of main tasks to do, aka grouping tasks by user interests (categories)

Proposed dashboard categories: [edited to integrate max comment below]

  - `merge` : entities could be of any type. 
    - `works` : already developed
    - `humans`: should first return all user feedback tasks (aka "reporter tasks"), then autogenerated tasks
  - `delete` : entities could be of any type
    - `works`
    - `humans`

Proposed implementation:
- add a `delete` task type, aside `deduplicate` (which could be renamed to `merge` (?))
- rename `by-entities-type` -> `by-type`: endpoint would have two arguments: `type` and `entities-type`. To be able to query `action=by-type?type=deduplicate&entities-type=humans` which would return only tasks with a reporter
- query autogenerated tasks would be kept as is (with `by-score` endpoint)
- since an entity would possibly both have a `merge` and a `delete` task, endpoint `by-suspect-uri` would necessarily need a `type` argument.
- rename `suspectUri` -> `uri` (as `suspect` term does not qualify anything useful)

maxlath:
What do you think of unifying user generated and robot generated merge tasks? The current human deduplication process would just be one provider among others of merge tasks, specialized on humans, but the rest would be entity-type agnostic and reporter (user or bot) agnostic(?). In that direction, and working from memories, I think it might make sense to make that author deduplication process create less tasks: it automerges what it can and creates tasks when it's not quite sure, but doesn't create a task for every homonym returned by Elasticsearch(?) as that's of a lower information quality

jum-s:
Here is a query to find the 1000th task sorted by descending score
```
curljson "http://[couchdb]/tasks-prod/_design/tasks/_view/byScore" | jq ".rows[1:768396]|sort_by(.key[2])|reverse|.[10000]"
``` 

(`[1:768396]` is hack to make the dump parsable by jq)

This could allow us to set a threshold of `350`, and still create 15k tasks (against 700k today). Threshold could be a config setting to allow to recreate tasks without having to push a commit.

[...]

PR #735, opened 2024-02-28
Task reorganise deduplicates

jum-s:
This PR is a preparation for tasks related issues (namely #724 and #491) which will introduce another `type` (in task document), namely `delete`. This addition will involve several API changes, which this PR tries to address:

- rename endpoint `by-type` -> `by-entities-type` and move parameter and `type` to `entities-type` see commit named "add a task type parameter"
- add a `minimumScoreToAutogenerate` config parameter to not auto generate too many entities (see 97a0e8980)

Client PR: https://github.com/inventaire/inventaire-client/pull/503


jum-s:
Rebasing after tasks have been "typscripted" on `main` was long, but should be okay for merge now.

server/controllers/tasks/lib/tasks.ts
```diff
@@ -82,20 +84,29 @@ export function getTasksBySuggestionUri (suggestionUri) {
   return db.getDocsByViewKey<Task>('bySuggestionUriAndState', [ suggestionUri, null ])
 }
 
-export async function getTasksBySuspectUris (suspectUris: EntityUri[], options: TasksQueryOptions = {}) {
+export async function getTasksBySuspectUrisAndType (uris: EntityUri[], types: string[]) {
+  const keys = zip(uris, types)
```
maxlath, line 88: I don't think that works for cases where `uris.length !== 1 || types.length !== 1`
```js
zip([uriA, uriB, uriC], ['deduplicate'],)
// => [ [ uriA, 'deduplicate' ], [ uriB, undefined ], [ uriC, undefined ] ]
```

```suggestion
  const keys = types.flatMap(type => uris.map(uri => [ uri, type ]))
```

tests/api/tasks/by_entities_uris.test.ts
```diff
@@ -1,60 +1,14 @@
 import 'should'
```
maxlath, line 1: When running these tests with empty databases (non-quick), I get the 3 following red tests:

```
  1) tasks:byEntitiesType
       should return tasks with a specific entities type:
     TypeError: Cannot read properties of undefined (reading 'entitiesType')
      at Context.<anonymous> (tests/api/tasks/by_entities_type.test.ts:34:14)

  2) tasks:byEntitiesType
       should return a limited array of tasks:

      AssertionError: expected 0 to be 1
      + expected - actual

      -0
      +1
      
      at Assertion.fail (node_modules/.pnpm/should@13.2.3/node_modules/should/cjs/should.js:275:17)
      at Assertion.value (node_modules/.pnpm/should@13.2.3/node_modules/should/cjs/should.js:356:19)
      at Context.<anonymous> (tests/api/tasks/by_entities_type.test.ts:40:25)

  3) tasks:byEntitiesType
       should take an offset parameter:
     TypeError: Cannot read properties of undefined (reading 'should')
      at Context.<anonymous> (tests/api/tasks/by_entities_type.test.ts:47:15)

```

server/controllers/tasks/lib/tasks.ts
```diff
@@ -108,7 +119,7 @@ function getKeys (uris: EntityUri[], includeArchived?: boolean) {
 
 const buildKey = state => uri => [ uri, state ]
 
-function completeWithEmptyArrays (getTasksByUris, uris: EntityUri[]) {
+const fillWithEmptyArrays = (getTasksByUris, uris: EntityUri[]) => {
```
maxlath, line 122: [non-blocking]
The stylistic trend I follow is rather transitioning the other way around (from arrow functions to named functions), outside of callbacks

server/controllers/tasks/tasks.ts
```diff
@@ -25,8 +25,10 @@ export default {
       'deduplicate-works': deduplicateWorks,
     },
     admin: {
-      'collect-entities': collectEntities,
-      'check-entities': checkEntities,
+      // Untested endpoint which triggers a global check of local database entities,
```
maxlath, line 28: It has been tested, just not automatically. The current wording might make it sound like it has never even been tried ^^'
```suggestion
      // Endpoint with no automatic tests coverage that triggers a global check of local database entities,
```

tests/api/fixtures/tasks.ts
```diff
@@ -1,6 +1,5 @@
-import { map } from 'lodash-es'
 import { createTasksInBulk } from '#controllers/tasks/lib/tasks'
-import { checkEntities } from '../utils/tasks.js'
```
maxlath, line 3: `createSomeTasks` still seems to use those dependencies, but since 43aa8b7 there are no consumer for `createSomeTasks` so it can also be removed I guess

tests/api/fixtures/tasks.ts
```diff
@@ -21,8 +20,10 @@ export function createSomeTasks (humanLabel) {
 
 export async function createTask (params) {
   const taskDoc = await createTaskDoc(params)
-  return createTasks([ taskDoc ])
+  const taskRes = await createTasks([ taskDoc ])
   .then(tasks => tasks[0])
```
maxlath, line 24: [non-blocking]
```suggestion
  const [ taskRes ] = await createTasks([ taskDoc ])
```

tests/api/tasks/by_entities_uris.test.ts
```diff
@@ -67,11 +21,19 @@ describe('tasks:bySuspectUris', () => {
     tasks[uri][0].should.be.an.Object()
   })
 
+  it('with empty type should return an array of tasks', async () => {
```
maxlath, line 24: [non-blocking] Keeping the "it should" form, unless too farfetched, reduces brain hick ups ;)
```suggestion
  it('should return an array of tasks when passed an empty type ', async () => {
```

server/controllers/tasks/lib/get_new_suggestions_or_automerge.ts
```diff
@@ -0,0 +1,169 @@
...
+import { prefixifyInv } from '#controllers/entities/lib/prefix'
+import { propertiesValuesConstraints as properties } from '#controllers/entities/lib/properties/properties_values_constraints'
+import { getEntityNormalizedTerms } from '#controllers/entities/lib/terms_normalization'
+import typeSearch from '#controllers/search/lib/type_search'
+import { isNonEmptyString } from '#lib/boolean_validations'
+import { forceArray, someMatch } from '#lib/utils/base'
+import { automerge, validateAndAutomerge } from './automerge.js'
+import { findAuthorWithMatchingIsbnInWikipediaArticles } from './find_authors_with_isbns_in_wikipedia_articles.js'
+
+const { minimumScoreToAutogenerate } = CONFIG.tasks
```
maxlath, line 17: [non-blocking]
Does that really need to be configurable? Looks like a hard coded value, that we might want to adjust from time to time, would do just fine

maxlath (commented):
Hooray generalization! :tada:

server/controllers/tasks/tasks.ts
```diff
@@ -1,21 +1,21 @@
 import { initTasksHooks } from '#controllers/tasks/hooks'
 import ActionsControllers from '#lib/actions_controllers'
-import byEntitiesType from './by_entities_type.js'
-import { bySuspectUris, bySuggestionUris } from './by_entity_uris.js'
-import byIds from './by_ids.js'
-import byScore from './by_score.js'
-import checkEntities from './check_entities.js'
-import collectEntities from './collect_entities.js'
-import deduplicateWorks from './deduplicate_works.js'
-import update from './update.js'
```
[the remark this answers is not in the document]
jum-s: done in b9662a9bd

tests/api/tasks/by_entities_uris.test.ts
jum-s: done in a2525a960

tests/api/fixtures/tasks.ts
jum-s: done in 7d0a14873

tests/api/fixtures/tasks.ts
jum-s: done in 7d0a14873

server/controllers/tasks/tasks.ts
jum-s: done in e620dc3691

server/controllers/tasks/lib/tasks.ts
jum-s: fixed in 8c45d0cab

tests/api/tasks/by_entities_uris.test.ts
jum-s: Fixed in 7d0a14873 
Fixtures created human tasks instead instead of work tasks.

server/controllers/tasks/lib/tasks.ts
jum-s: fixed in a2525a9602

server/controllers/tasks/tasks.ts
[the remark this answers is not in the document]
jum-s: done in e620dc369

server/controllers/tasks/lib/get_new_suggestions_or_automerge.ts
maxlath: Let's roll with that then

tests/api/tasks/by_entities_uris.test.ts
maxlath: One red test remaining (with or without empty databases):
```
    1) should return tasks with a specific entities type


  0 passing (199ms)
  1 failing

  1) tasks:byEntitiesType
       should return tasks with a specific entities type:
     TypeError: Cannot read properties of undefined (reading 'entitiesType')
      at Context.<anonymous> (tests/api/tasks/by_entities_type.test.ts:35:14)
```
The `tasks` array is empty

jum-s: Some `await` was missing,
should be fixed now in 2a80923ff

maxlath (changes requested):
Several problems remain:
- The rename from `server/controllers/tasks/lib/get_new_tasks.ts` to `server/controllers/tasks/lib/get_new_suggestions_or_automerge.ts`  wasn't properly done, as that initial file wasn't deleted (so we don't see a diff but a file creation)
- Same with `server/controllers/tasks/lib/check_entity.ts` renamed to `server/controllers/tasks/lib/check_human_duplicate.ts` with the added twist that some changes were later done on both those files:
  - `check_entity.ts` was updated to `import getNewSuggestionsOrAutomerge from './get_new_suggestions_or_automerge.js'`, while `check_human_duplicate.js` still `import getNewTasks from './get_new_tasks.js'`. When correcting that, the `should create tasks for the requested URIs` test fails.
  
  I tried to fix that with properly amended commits (so that we preserve the `git log/diff --follow` features), but as a test was failing with the end result, I didn't dare to force push the result. I pushed it on branch `task-reorganise-duplicates-max` instead.

jum-s:
> When correcting that, the `should create tasks for the requested URIs` test fails.

When i correct this, the tests are green (quick and non-quick), on your branch too. If i delete task db, works too. Is there a possibility that the entities-test db having some format problem?

maxlath (approved):

maxlath:
So now the tests pass, but I can't tell what changed :sweat_smile: Lets merge this

PR #737, opened 2024-03-19
Tests cleanup

jum-s:
to be merged after #706

### DIFF
--- a/tests/api/entities/publisher_publications.test.ts
+++ b/tests/api/entities/publisher_publications.test.ts
@@ -67,8 +67,7 @@ describe('entities:publisher-publications', () => {
   })
 
   it('should get wikidata publisher collections', async () => {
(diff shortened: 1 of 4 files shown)


tests/api/entities/publisher_publications.test.ts
```diff
@@ -67,8 +67,7 @@ describe('entities:publisher-publications', () => {
   })
 
   it('should get wikidata publisher collections', async () => {
-    const { collections, editions } = await publicReq('get', `${endpoint}&uri=wd:Q2823584`)
-    editions.should.deepEqual([])
```
maxlath, line 71: This test assumes that no local edition was created with `wdt:P123=wd:Q2823584`, and that the SPARQL request to query.wikidata.org does not return Wikidata editions (in line with [wd editions quarantine](https://github.com/inventaire/inventaire/issues/182)). If you local test finds results, are those results local entities or wikidata?

PR #751, opened 2024-06-26
Listing element comment

jum-s:
Should be rebased and merged after #677 is merged on main.

Most of the work is [client sided](https://github.com/inventaire/inventaire-client/pull/501)

### DIFF
--- a/server/controllers/listings/by_element_id.ts
+++ b/server/controllers/listings/by_element_id.ts
@@ -1,0 +1,25 @@
+import { property } from 'lodash-es'
+import { getElementById } from '#controllers/listings/lib/elements'
+import { getListingById } from '#controllers/listings/lib/listings'
+import { unauthorizedError } from '#lib/error/error'
+import { filterVisibleDocs } from '#lib/visibility/filter_visible_docs'
+import type { ListingElement } from '#types/element'
+
+const sanitization = {
+  id: {},
+}
+
+async function controller ({ id, req, reqUserId }) {
+  const element: ListingElement = await getElementById(id)
+  const listing = await getListingById(element.list)
+  const allVisibleListings = await filterVisibleDocs([listing], reqUserId)
+  if (!allVisibleListings.map(property('_id')).includes(element.list)) {
+    throw unauthorizedError(req, 'unauthorized access', { id })
+  }
+  return {
+    element,
+    list: listing,
+  }
+}
+
+export default { sanitization, controller }
--- a/server/controllers/listings/by_element_id.ts
+++ b/server/controllers/listings/by_element_id.ts
@@ -1,7 +1,6 @@
-import { property } from 'lodash-es'
 import { getElementById } from '#controllers/listings/lib/elements'
 import { getListingById } from '#controllers/listings/lib/listings'
-import { unauthorizedError } from '#lib/error/error'
+import { newError, notFoundError } from '#lib/error/error'
 import { filterVisibleDocs } from '#lib/visibility/filter_visible_docs'
 import type { ListingElement } from '#types/element'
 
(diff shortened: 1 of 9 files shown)


server/models/element.ts
```diff
@@ -53,13 +65,29 @@ export function updateElementDoc (newAttributes, oldElement) {
...
     }
-    const newVal = newAttributes[attribute]
+    let newVal
+    if (attribute === 'ordinal') {
+      newVal = findNewOrdinal(oldElement, listingElements, newAttributes[attribute])
+    } else {
+      newVal = newAttributes[attribute] || defaultValues[attribute]?.()
+      validations.pass(attribute, newVal)
+    }
+    if (newVal !== undefined) {
```
maxlath, line 75: [non-blocking]
The convention within this code base is to check against both `null` and `undefined`, with this Douglas Crockford-approved exception use of the non-strict-egal comparison:
 ```suggestion
     if (newVal != null) {
```
which is strictly equivalent to 
 ```js
    if (newVal !== undefined && newVal !== null) {
```

server/controllers/listings/by_element_id.ts
```diff
@@ -0,0 +1,25 @@
...
+
+const sanitization = {
+  id: {},
+}
+
+async function controller ({ id, req, reqUserId }) {
+  const element: ListingElement = await getElementById(id)
+  const listing = await getListingById(element.list)
+  const allVisibleListings = await filterVisibleDocs([listing], reqUserId)
+  if (!allVisibleListings.map(property('_id')).includes(element.list)) {
```
maxlath, line 16: [non-blocking]
As we are checking a single listing, this can be much simpler
```suggestion
  const visibleListings = await filterVisibleDocs([listing], reqUserId)
  if (visibleListings.length === 0) {
```

server/controllers/listings/by_element_id.ts
```diff
@@ -0,0 +1,25 @@
...
+import { getListingById } from '#controllers/listings/lib/listings'
+import { unauthorizedError } from '#lib/error/error'
+import { filterVisibleDocs } from '#lib/visibility/filter_visible_docs'
+import type { ListingElement } from '#types/element'
+
+const sanitization = {
+  id: {},
+}
+
+async function controller ({ id, req, reqUserId }) {
```
maxlath, line 12: It would be nice to have a couple of tests for this controller (in particular: what if we request a non-existing element, a non-authorized element). It would have revealed that `req` is not defined on the controller params object, but as second argument.

```suggestion
async function controller ({ id, reqUserId }, req) {
```

But we actually don't need to access it here, see below.

server/controllers/listings/by_element_id.ts
```diff
@@ -0,0 +1,25 @@
...
+  const element: ListingElement = await getElementById(id)
+  const listing = await getListingById(element.list)
+  const allVisibleListings = await filterVisibleDocs([listing], reqUserId)
+  if (!allVisibleListings.map(property('_id')).includes(element.list)) {
+    throw unauthorizedError(req, 'unauthorized access', { id })
```
maxlath, line 17: Thank to the `ActionsControllers` wrapper, we don't need to handle `req`
```suggestion
    throw newError('unauthorized access', 403, { id })
```

jum-s: done in 08fdfab86

server/controllers/listings/by_element_id.ts
jum-s: done in 08fdfab86

server/models/element.ts
jum-s: done in 08fdfab86

server/controllers/listings/by_element_id.ts
jum-s: I added some tests in 08fdfab86 which logically fixed the bugs you mentioned

Issue #716, opened 2024-01-26
Display list with a set entities type


jum-s:
At list creation, one may select an entity type (default to 'works or series', then 'authors' and 'publishers') that would enable a typed search for elements

jum-s:
Allowing more entities types may come with a better way to add elements to a list. In order to provide some entities suggestions (aka on click of the search textfield)

- reuse `worksTree` entities and display as default results in the autocomplete
- OR click on a button "Add from your inventory", next to the autocomplete which would display a hardly modified `ShelfItemsAdder` modal

PR #752, opened 2024-06-28
Listing entities type

jum-s:
Add a `type` key (maybe `entitiesType` is more DSL friendly?) on Listing model.

Solves issue #716
client PR [502](https://github.com/inventaire/inventaire-client/pull/502)

### DIFF
--- a/server/models/attributes/listing.ts
+++ b/server/models/attributes/listing.ts
@@ -10,5 +10,8 @@ export default {
     'creator',
     'type',
   ]),
-  type: [ 'work' ],
+  type: [ 'work', 'serie' ],
+  entityTypesByListingType: {
+    work: [ 'work', 'serie' ],
+  },
 }
--- a/server/models/attributes/listing.ts
+++ b/server/models/attributes/listing.ts
@@ -8,5 +8,7 @@ export default {
   updatable,
   validAtCreation: updatable.concat([
     'creator',
+    'type',
   ]),
+  type: [ 'work' ],
 }
--- a/server/models/attributes/listing.ts
+++ b/server/models/attributes/listing.ts
@@ -4,16 +4,18 @@ const updatable = [
   'name',
 ]
 
+const entityTypesByListingType = {
+  work: [ 'work', 'serie' ],
+  author: [ 'human' ],
+  publisher: [ 'publisher' ],
+}
+
 export default {
   updatable,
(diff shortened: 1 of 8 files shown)


server/models/attributes/listing.ts
```diff
@@ -8,5 +8,12 @@ export default {
   updatable,
   validAtCreation: updatable.concat([
     'creator',
+    'type',
   ]),
+  type: [ 'work', 'author', 'publisher' ],
```
maxlath, line 13: What about keeping `entityTypesByListingType` as the single source of truth for listing types?
```suggestion
  type: Object.keys(entityTypesByListingType),
```

server/models/listing.ts
```diff
@@ -10,6 +10,9 @@ export function createListingDoc (listing) {
   assert_.string(listing.creator)
   assert_.string(listing.name)
 
+  if (!listing.type) {
+    listing.type = 'work'
+  }
```
maxlath, line 15: [non-blocking]
```suggestion
  listing.type ??= 'work'
```

server/types/listing.ts
```diff
@@ -7,6 +7,7 @@ export type ListingId = CouchUuid
 
 export interface Listing extends CouchDoc {
   _id: ListingId
+  type: string
```
maxlath, line 10:
```suggestion
  type: keyof typeof entityTypesByListingType
```
Having a more narrow types will help with things such as detection of the missing/invalid per-type i18n keys  client side

maxlath (commented):
I think keeping `type` should be fine, especially if the possible values differ from the `entity.type` values (ex: "author" instead of "human", or "work" for both "work" and "series")

maxlath:
Required listing docs migration post-merge:
```js
export default function (listing) {
  listing.type = 'work'
  return listing
}
```

server/models/attributes/listing.ts
jum-s: done in 885ec0428

server/models/listing.ts
jum-s: done in 155a03b71

server/types/listing.ts
jum-s: done in ed0a6b78

jum-s:
Rebasing should not conflict, if okay to merge

Issue #192, opened 2018-06-14
remove ISBNs status of canonical edition URIs ids


maxlath:
preferring `inv` URIs over `isbn` URIs would have several advantages:
* simplify entities code: by turning ISBN URIs into aliases among other ids, that would let URIs to be either `inv` or `wd`, making many things simpler (merge, revert merge, entities and items redirections, etc)
* that would also open the possibility to have several editions with the same ISBN, in cases where the publisher didn't care to observe the 1 edition = 1 ISBN relation rule

maxlath:
examples of several editions for one ISBN:
* [978-2-9534848-1-6](https://br.wikipedia.org/wiki/Sav-Heol#Levrio%C3%B9_deskarded_pe_studi) (spotted by @belett, which confirmed with the books at hand that it isn't a mistake on Wikipedia >< )

belett:
In fact, there is (at least) two different situations that broke the 1 edition = 1 ISBN :
- an edition with two corrects ISBN, very rare but it can happen (especially for the 'beaux livres"), "Histoire de Rennes" for instance is printed by two editors each giving an ISBN : 978-2-84398-378-8 for Apogée and  ISBN 978-2-7535-1260-3 for the Presses universitaires de Rennes
- an edition with an wrong ISBN printed in it and then a correct ISBN in the database (also rare but common for small publishers), here this is the case mentioned, ISBN 978-2-9534848-1-6 is wrongly printed in the back but ISBN 978-2-9534848-3-0 is the correct one.

The first case is weird but the second case is the most trickiest because when someone has the book in them hands, there is no way to know that the ISBN is wrong.

jum-s:
2 editors and 2 isbns on one edition, what a funny world we live in..
The "wrong" isbn case happens because of database attempts defined before printing. This *a priori* definition do not fit the physical world. Can we then say the printed isbn actually is the correct one ? Systems that use code bar scanners would. Human work sourcing and/or editors database sourcing.

maxlath:
maybe we need to introduce Wikidata-like references in Inventaire data model to say that the cover says something and the database something else ;)

belett:
> Can we then say the printed isbn actually is the correct one ?

In theory, we could.
In some cases that could work but most of the time, this would be **weird and dangerous**.

Here some examples:
[...]

Issue #181, opened 2018-05-22
recover a shadow entity feature


maxlath:
having the possibility to extend a Wikidata entity from Inventaire could be useful in some cases (#182).
We had that feature, but it was removed as it wasn't used defc9f7; it could just be recovered

jum-s:
I have not set my mind upon this shadow item thing, seems handy to define customed props ok. 

But, if in a federated world, other people can define the same claim with another value. What if inventaire shadow claim is a doublon of the wikidata instance with a misspelled value ? 
How do we deal with claims that could fall in a conflict with womeone else definition ?
Constraining those shades/claims to a few specific properties seems the least we can do (`P18: image`) 
In those experimental days of federation understanding, opening the shadows possibility might look like a slipery way, we dont really know where it could lead yet.

maxlath:
I agree, we should keep it to strictly necessary claims, like book covers that aren't supported by Wikidata, at least at first

maxlath:
A comment I made today in the chat on the matter:

> The tight integration [with Wikidata] indeed comes at a real cost on our nerves, as any Wikidata contributor can decide that a work entity is actually an edition and break things on our side, any Wikidata oddity becomes our problem.
> I have been playing with ideas that I hope would make us more resilient, such as having an entity document locally in Inventaire database for each Wikidata entity of interest, with which it would have the equivalent of a sameAs relation. That local document could then have claims such as a type lock (a way of saying "this entity should be considered a work, whatever its P31 now says"), and data that can't be stored in Wikidata (the typical use case being a book cover)

maxlath:
Instead of "shadow entity", I like the idea of an entity layer: several entity documents in several databases, identified as refering to the same concept ([`sameAs` relation](https://en.wikipedia.org/wiki/SameAs)), could be considered layers of the same entity. Such layers could overlap, or not. (In our case, I think they wouldn't need to)

maxlath:
Coming back on this issue, it seems clear that having a local layer with only local properties would be a safe first implementation: there does not seem to be a need to override wikidata claims, so no need to handle having local and remove claims for the same wikidata property.

maxlath:
Done in https://github.com/inventaire/inventaire/pull/753 :tada:

PR #753, opened 2024-07-28
Introduce local entity layers and remove properly shaped Wikidata editions from quarantine

maxlath:
Addresses #181 and #182. I explored both matters together as integrating Wikidata editions are the primary motivations to implement entity local layers, so both issues informed each other's implementation.

Client side PR https://github.com/inventaire/inventaire-client/pull/505

## Entity local layers
This PR introduces remote entities local layers (primarily in b695d77ac), that is: a document in the local `entities` database, with `.claims['invp:P1'] = [ wdUri ]`, where `invp:P1` is the "same as remote entity" property, and `wdUri` is the Wikidata entity uri. That local layer can only have local property claims (`invp:Px`).

Consequently, when a `GET /api/entities?action=by-uris&uris=${wdUri}` request is made, the response aggregates both the remote and the local claims. The returned serialized entities now include `wdId` and/or `invId` to let the client now about the different layers (use-case exemple: to generate Wikidata wiki or history URLs), even when not requesting the entities claims.

Several controllers were affected by those changes, in particular:
- Entities merge logic needed to be adapted to handle local layers, see 3507fc785 and e57b4d8ec
- `GET /api/entities?action=history` needed to be made to accept a `uri`, so that local patches could be requested for a remote entity 97c343939

## Remove properly shaped Wikidata editions from quarantine
Thanks to entity local layers, it's possible to use Wikidata editions entities with locally stored book covers (`invp:P2`). The quarantine was lifted in 0fe134528. Remain ignored Wikidata editions that have no value for `wdt:P629`, as those entities:
- lack that work-edition relation assumed to exist throughout the Inventaire code base
- are more likely to be later transformed into work entities

This PR proposes to also use `isbn:` prefixed uris as canonical uris for Wikidata editions, as, despite their flaws #192 , ISBNs are more stable than Wikidata edition entities, where contributors edit conflicts might disrupt the association between a wd entity and a given ISBN.

One consequence of including Wikidata editions is that when an entity is requested by ISBN (which is a hot path), we now need to lookup if there is an existing edition with that ISBN in Wikidata. This is done in bulk by `getWdEntitiesByIsbns`, which also caches the result.

Another consequence is that the mechanism to fetch data from external databases (typically national libraries) is not used any more for editions found on Wikidata. A future development could thus be to enable enriching Wikidata editions with data found in external databases, but that would require having a Wikidata user account for the Inventaire server, to sign those edits on Wikidata. 

### After merge
- [ ] Reconcile existing inv and wd edition entities

### Later
- [ ] Enable transfering inv editions to Wikidata


server/controllers/entities/lib/get_entities_by_isbns.ts
```diff
@@ -3,34 +3,44 @@ import { getInvEntitiesByIsbns } from '#controllers/entities/lib/entities'
...
+  // This should ideally be corrected to `SerializedEntity[] | ExpandedSerializedEntity[]` but that causes a type error
+  entities: (SerializedEntity | ExpandedSerializedEntity)[]
   notFound?: IsbnEntityUri[]
 }
 
 export async function getEntitiesByIsbns (rawIsbns: Isbn[], params: EntitiesGetterParams = {}) {
-  const [ isbns, redirections ] = getRedirections(rawIsbns)
+  const { isbns13h: isbns, redirections, parsedIsbnsData } = getIsbnsData(rawIsbns)
   const { autocreate, refresh } = params
+  const { entities: serializedWdEntities } = await getWdEntitiesByIsbns(parsedIsbnsData, params)
```
jum-s, line 24: Its quite a piece this `getEntitiesByIsbns` now, but if i get it well, there is a case that seem problematic to me:
when an inv entity with a given isbn, let's say an edition with claims carefully crafted over the years by inv users, also have a wikidata item with the same isbn, whats happening ? The wikidata edition will be returned, ignoring the inv entity (right?) 
If this is the case, that means some editions data will be displayed differently (ie. merging this as is will change some titles to be a wd item title, instead of the inv edition now). Wouldn't it be safer (and more respectful of inv user contributions) to give inv editions priority over wd editions?

maxlath: That's why there is the after-merge TODO to "reconcile existing inv and wd edition entities", which would include transferring the carefully crafted claims to Wikidata. After that, the inv entity would become a redirection, or, if it has a book cover (`invp:P2`), a local entity layer.

Good point on the title: maybe `wdt:P1476` should be a required property to consider a Wikidata edition as acceptable, just like `wdt:P629`.

jum-s: A more precise illustration in the same shape as my previous comment:
As a long term inv user, I crafted my series on inventaire some times ago with the great `/cleanup` tool, formatted the titles etc. Then, after this PR merge, my inventory will look like having different title formats for the same series I crafted.

About moving inv editions to wikidata, what if after the inv entity is becoming a redirection, then the wd edition becomes a work?

It feels Inventaire is going all the way from not accepting any wd edition to give priority to all wd editions. By leaving priority to inventaire community in this PR, we would ensure some minimal moves (roadmap is tight), and then switch to wikidata editions fully at the time the bulk redirection is implemented?

maxlath: >  Then, after this PR merge, my inventory will look like having different title formats for the same series I crafted.

It should not be the case, as inv title conventions are based on wd conventions.
Additionally, it will most likely be the case that your series do not have a matching edition in Wikidata yet: in the latest inv dump, there were 1030831 editions with an ISBN, while there are only [80327 editions with an ISBN in Wikidata (at least that we would consider valid)](https://query.wikidata.org/#SELECT%20%3Fedition%20%3Fisbn13h%20%3Fisbn10h%20WHERE%20%7B%0A%20%20%3Fedition%20wdt%3AP31%20wd%3AQ3331189%20.%0A%20%20%7B%20%3Fedition%20wdt%3AP212%20%3Fisbn13h%20.%20%7D%20UNION%20%7B%20%3Fedition%20wdt%3AP957%20%3Fisbn10h%20.%20%7D%0A%20%20FILTER%20EXISTS%20%7B%0A%20%20%20%20%3Fedition%20wdt%3AP629%20%3Fwork%20.%0A%20%20%20%20%3Fedition%20wdt%3AP1476%20%3Ftitle%20.%0A%20%20%7D%0A%7D). Furthermore, the [overlap with the inv editions with an ISBN, still based on this latest inv dump, concerns only 12005 editions](https://dumps.inventaire.io/tmp/wd_inv_editions_with_isbn_overlap.ndjson). I would expect a small minority of those to be conflicting.

> what if after the inv entity is becoming a redirection, then the wd edition becomes a work?

That's unfortunately a very legitimate concern from our experience, but my hope is that only accepting editions that already have an associated work (`wdt:P629`) would make this occurrence quite rare: most cases of such a type change seem to occur for isolated "book" items.

maxlath: @jum-s  I inverted the priorities in 1532b2dce: existing local entities should now be found first, until they are merged with the remote entity.

jum-s: Could it be rebased (there are some conflicts) so that couch doesnt have to reindex all entities?

maxlath: @jum-s rebased

tests/api/entities/get/by_isbns.test.ts
```diff
@@ -15,6 +19,64 @@ describe('entities:get:by-isbns', () => {
     should(res.notFound).not.be.ok()
   })
 
+  describe('wikidata editions', () => {
+    it('should find Wikidata edition entities by its isbn uri', async () => {
```
jum-s, line 23: The test is (the only) red.

```
     TypeError: Cannot read properties of undefined (reading 'should')
      at Context.<anonymous> (tests/api/entities/get/by_isbns.test.ts:28:32)
```

Is something wrong with my config ? Is it green on your side ?
The data i have is quite new (ie. cache was busted)

maxlath: I pushed a debug commit: can you fetch, rerun this test and report the server and tests log lines starting with "🚀"?

maxlath: > logs above shows there is an inv local entity with this ISBN, which I guess you don't have in your local db

I assumed you tried with an empty `entities-tests` database: yes, if there is already a local entity with that ISBN, that test will fail due to the priority now being given to the local entity (since 9fb9a80cb). Can you retry with an empty  `entities-tests`  db?

jum-s: 'invp:P1' is a property based on same as entities only. Well, those comments helped to understand the codebase. Thanks!

jum-s (approved):

Issue #704, opened 2023-11-27
make publisher collections transferable to wikidata


maxlath:
Currently, collections might look something like this:

```js
{
  labels: {},
  claims: {
    'wdt:P31': [ 'wd:Q20655472' ],
    'wdt:P1476': [ 'Nautilus Flugschrift' ],
    'wdt:P123': [ 'wd:Q1285306' ]
  }
}
```
The title (`wdt:P1476`) is a simple string, without the language code expected by [`wikibase-edit`](https://github.com/maxlath/wikibase-edit/blob/main/docs/how_to.md#incremental-mode) for `monolingualtext` properties such as [`wdt:P1476`](https://www.wikidata.org/wiki/Property:P1476):
```js
    'wdt:P1476': [ { text: 'Nautilus Flugschrift', language: 'de' } ],
```

In [other places](https://github.com/inventaire/inventaire-client/blob/2d56337/app/modules/entities/components/editor/lib/create_helpers.js#L35-L39), the language code is inferred from the language claim (`wdt:P407`). The same could be done for collections being transferred to wikidata:
* enable `wdt:P407` for entities of type `collection`
* when transferring to wikidata:
  * validate that the collection has a `wdt:P407` claim set
  * deduce the language code from the `wdt:P407` claim, to propertly reshape the `monolingualtext` value

maxlath:
Done in https://github.com/inventaire/inventaire/pull/760 and https://github.com/inventaire/inventaire-client/pull/507

PR #760, opened 2024-09-04
allow moving editions and publisher collections to wikidata

maxlath:
Addresses #704 

Client PR: https://github.com/inventaire/inventaire-client/pull/507

TODO after merge:
* [ ] clear outdated SPARQL queries cache: `lev ./db/leveldb-prod --prefix '!cache!wdQuery:resolve_external_ids' --del | lev ./db/leveldb-prod --batch`

### DIFF
--- a/server/controllers/entities/lib/create_wd_entity.ts
+++ b/server/controllers/entities/lib/create_wd_entity.ts
@@ -1,17 +1,20 @@
-import { getClaimValue } from '#controllers/entities/lib/inv_claims_utils'
+import { getClaimValue, getFirstClaimValue } from '#controllers/entities/lib/inv_claims_utils'
 import { getWikidataOAuthCredentials, validateWikidataOAuth } from '#controllers/entities/lib/wikidata_oauth'
 import { newError } from '#lib/error/error'
 import { mapKeysValues, objectEntries } from '#lib/utils/base'
+import { requireJson } from '#lib/utils/json'
 import { log } from '#lib/utils/logs'
 import { relocateQualifierProperties } from '#lib/wikidata/data_model_adapter'
 import wdEdit from '#lib/wikidata/edit'
-import type { EntityUri, EntityValue, ExpandedClaims, InvExpandedPropertyClaims, InvSnakValue, Labels, PropertyUri, Reference, ReferenceProperty, ReferencePropertySnaks, WdEntityId, WdEntityUri, WdPropertyId } from '#server/types/entity'
+import type { EntityUri, EntityValue, ExpandedClaims, InvExpandedPropertyClaims, InvSnakValue, Labels, PropertyUri, Reference, ReferenceProperty, ReferencePropertySnaks, WdEntityId, WdEntityUri, WdPropertyId, InvClaimObject } from '#server/types/entity'
 import type { User } from '#server/types/user'
 import { getInvEntityType } from './get_entity_type.js'
 import { prefixifyWd, unprefixify } from './prefix.js'
 import { getPropertyDatatype } from './properties/properties_values_constraints.js'
 import { validateInvEntity } from './validate_entity.js'
-import type { SimplifiedQualifiers } from 'wikibase-sdk'
+import type { SimplifiedQualifiers, WikimediaLanguageCode } from 'wikibase-sdk'
+
+const wmLanguageCodeByWdId = requireJson('wikidata-lang/mappings/wm_code_by_wd_id.json')
 
 const allowlistedEntityTypes = [ 'work', 'serie', 'human', 'publisher', 'collection' ]
 
@@ -93,6 +96,7 @@ function format (entity: EntityDraft) {
   // Relocate qualifier properties after unprefixifying,
   // as the unprefixifyClaims function doesn't handle qualifiers
   relocateQualifierProperties(entity)
+  reshapeMonolingualTextClaims(entity)
   return entity
 }
 
@@ -122,4 +126,37 @@ function unprefixifySnakValues (property: ReferenceProperty, propertyValues: Ref
   } else {
     return propertyValues
   }
+}
+
+const monolingualProperties = [
+  'wdt:P1476',
+  'wdt:P1680',
+] as const satisfies PropertyUri[]
+
+function reshapeMonolingualTextClaims (entity: EntityDraft) {
+  const { claims } = entity
+  if (!monolingualProperties.find(property => claims[property])) return
+  const languageUri = getFirstClaimValue(claims, 'wdt:P407') as WdEntityUri
+  if (!languageUri) {
+    throw newError('monolingual text claims can not be reshaped in absence of a language claim', 400, { entity })
+  }
+  const langWdId = unprefixify(languageUri)
+  const languageCode = wmLanguageCodeByWdId[langWdId] as WikimediaLanguageCode
+  if (!languageCode) {
+    throw newError('wikimedia language code not found', 400, { langWdId })
+  }
+  for (const property of monolingualProperties) {
+    reshapeMonolingualTextPropertyClaims(claims, property, languageCode)
+  }
+}
+
+function reshapeMonolingualTextPropertyClaims (claims: ExpandedClaims, property: PropertyUri, languageCode: WikimediaLanguageCode) {
+  if (!claims[property]) return
+  claims[property] = claims[property].map((claimObject: InvClaimObject) => {
+    const { value, references } = claimObject
+    return {
+      value: { text: value, language: languageCode },
+      references,
+    }
+  })
 }
--- a/server/controllers/entities/lib/create_wd_entity.ts
+++ b/server/controllers/entities/lib/create_wd_entity.ts
@@ -1,7 +1,7 @@
 import { getClaimValue, getFirstClaimValue } from '#controllers/entities/lib/inv_claims_utils'
 import { getWikidataOAuthCredentials, validateWikidataOAuth } from '#controllers/entities/lib/wikidata_oauth'
 import { newError } from '#lib/error/error'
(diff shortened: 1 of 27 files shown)


server/controllers/entities/lib/create_wd_entity.ts
```diff
@@ -1,17 +1,20 @@
...
 import { getInvEntityType } from './get_entity_type.js'
 import { prefixifyWd, unprefixify } from './prefix.js'
 import { getPropertyDatatype } from './properties/properties_values_constraints.js'
 import { validateInvEntity } from './validate_entity.js'
-import type { SimplifiedQualifiers } from 'wikibase-sdk'
+import type { SimplifiedQualifiers, WikimediaLanguageCode } from 'wikibase-sdk'
+
+const wmLanguageCodeByWdId = requireJson('wikidata-lang/mappings/wm_code_by_wd_id.json')
 
 const allowlistedEntityTypes = [ 'work', 'serie', 'human', 'publisher', 'collection' ]
```
jum-s, line 19: If this PR also intends to enable moving an edition to wikidata therefore:

```suggestion
const allowlistedEntityTypes = [ 'work', 'serie', 'human', 'publisher', 'collection', 'edition' ]
```

But then, there is also some work to tranfer the image `inv:P2`, and there might be some work to handdle the case where the isbn value is already used by a wd item. 

So maybe the feature "move an edition to wikidata" could be done in a separate PR (?)

server/controllers/entities/lib/move_to_wikidata.ts
```diff
@@ -24,9 +27,25 @@ export async function moveInvEntityToWikidata (user: User, invEntityUri: InvEnti
...
-  if ('claims' in entity) claims = extendInvClaims(entity.claims)
+  if ('claims' in entity) claims = expandInvClaims(entity.claims)
 
-  const { uri: wdEntityUri } = await createWdEntity({ labels, claims, user, isAlreadyValidated: true })
+  const conflictingWdEntities = await resolveExternalIds(claims, {
+    resolveOnWikidata: true,
+    resolveLocally: false,
+    refresh: true,
+  })
+  if (conflictingWdEntities.length > 0) {
```
jum-s, line 37: When I create an edition without isbn, then click on 'move to Wikidata' following error pops up :

```
****** Cannot read properties of undefined (reading 'length') ******
TypeError: Cannot read properties of undefined (reading 'length')
    at moveInvEntityToWikidata (server/controllers/entities/lib/move_to_wikidata.ts:37:29)
    at async controllerWrapper (server/lib/controller_wrapper.ts:33:22) {
[skip...]
  context: { body: { uri: 'inv:6d495495dee900063c5ee8afc349ed7f' } }
}
-----
PUT /api/entities?action=move-to-wikidata 500 10ms
```

and indeed `resolve_external_ids.ts` line 43:
```
if (propertyValuePairs.length === 0) return
```
returns an `undefined` value

```suggestion
  if (conflictingWdEntities && conflictingWdEntities.length > 0) {
```

jum-s: When i solve the previous bug above, another error pops up (still on edition's "move to wikidata"):

```
****** wd entity creation ******
{
  labels: {},
  claims: {
    'wdt:P1476': [ { value: 'La méthode Google' } ],
    'wdt:P577': [ { value: '2009' } ],
    'wdt:P1104': [ { value: 426 } ],
    'wdt:P2969': [ { value: '5100525' } ],
    'wdt:P31': [ { value: 'wd:Q3331189' } ],
    'wdt:P629': [ { value: 'wd:Q100989307' } ],
    'wdt:P212': [ { value: '978-0-06-170971-5' } ],
    'wdt:P957': [ { value: '0-06-170971-9' } ],
    'wdt:P407': [ { value: 'wd:Q1860' } ]
  }
}
-----
****** 400 ******
Error: invalid monolingualtext value
    at Object.new (inventaire/node_modules/wikibase-edit/lib/error.js:3:17)
    at Object.snakValue (inventaire/node_modules/wikibase-edit/lib/validate.js:76:23)
    at fullClaimBuilder (inventaire/node_modules/wikibase-edit/lib/entity/build_claim.js:52:14)
    at module.exports (inventaire/node_modules/wikibase-edit/lib/entity/build_claim.js:21:12) {
  statusCode: 400,
  context: { property: 'P1476', value: 'La méthode Google' },
  user: { _id: '5785b6c45d065900e48c865e0b00084e', username: 'adamsberg' },
  headers: {
    referer: 'http://localhost:3005/entity/isbn:9780061709715/edit'
  }
}
-----
PUT /api/entities?action=move-to-wikidata 400
```

[edit: when i click on move to wd from a collection editor, then I am asked to add a language claim (aka `This collection can not be moved to Wikidata as it lacks a language claim`, then i add a language claim, then when i click, i have the same error as above `invalid monolingualtext value`)]

server/controllers/entities/lib/resolver/resolve_edition.ts
```diff
@@ -1,33 +1,33 @@
...
-  .then(pickUriFromResolversResponses)
-  .then(uri => {
-    if (uri != null) { entry.edition.uri = uri }
-    return entry
-  })
+  // TODO: handle possible conflict between uriFoundByIsbn and urisFoundByExternalIds
+  let uri
+  if (uriFoundByIsbn) {
+    uri = uriFoundByIsbn
+  } else if (resultsFoundByExternalIds && resultsFoundByExternalIds.length === 1) {
```
jum-s, line 22:
```suggestion
  } else if (isNonEmptyArray(resultsFoundByExternalIds)) {
```

[non-blocking]

server/controllers/entities/lib/resolver/resolve_external_ids.ts
```diff
@@ -1,49 +1,81 @@
...
 
-export async function resolveExternalIds (claims: Claims, resolveOnWikidata = true) {
-  const externalIds = []
+interface ResolveExternalIdsOptions {
+  resolveOnWikidata?: boolean
+  resolveLocally?: boolean
+  refresh?: boolean
+}
+
+interface FormattedResult {
```
jum-s, line 20: A more context specific proposition (already in use in `resolve_edition.ts#L13`:

```suggestion
interface ResultsFoundByExternalIds {
```

server/controllers/entities/lib/move_to_wikidata.ts
maxlath: Some testing was indeed much needed! Fixed in 06939e532

server/controllers/entities/lib/resolver/resolve_edition.ts
maxlath: I think that `if (resultsFoundByExternalIds.length > 1)` it should not resolve, as we would not which of the found editions is the right one

server/controllers/entities/lib/resolver/resolve_external_ids.ts
maxlath: In that case, we have a more specific name at hand 6977b03eb

server/controllers/entities/lib/move_to_wikidata.ts
```diff
@@ -24,9 +27,25 @@ export async function moveInvEntityToWikidata (user: User, invEntityUri: InvEnti
...
+    resolveLocally: false,
+    refresh: true,
+  })
+  if (conflictingWdEntities?.length > 0) {
+    throw newError('Can not move to Wikidata: some Wikidata entities share the same identifiers', 400, { conflicts: conflictingWdEntities })
+  }
+
+  // Local claims will be preserved in a local layer during merge
+  const claimsWithoutLocalClaims = omitBy(claims, (propertyClaims, property) => isInvPropertyUri(property))
+  const { uri: wdEntityUri } = await createWdEntity({
```
jum-s, line 43: What about a `try...catch` around this function which touches an external service ? 
Otherwise, `mergeEntities` below will crash with some `undefined` `wdEntityUri` which may be poor for the nice admin who will review logs later on

server/lib/utils/base.ts
```diff
@@ -199,3 +199,7 @@ export function invert (obj: Record<string | number, string | number>) {
   }
   return invertedObj
 }
+
+export function getOptionalValue <T extends object, K extends keyof T> (obj: T, attribute: K) {
```
jum-s, line 203: [naming is hard] I'm confused by the `get` part (suggesting something more far-fetched than what it is (ie. a db request)) and by the `optional` part (isn't it always optional if the key is not defined in the object?)

Suggestions by order of preference:
- property : which gives a lodash/underscore flavour
- propertyValue: less generic than above
- objectValue
- objectKeyValue

I think not having a verb would be preferable to indicate that it's a small function

server/controllers/entities/lib/move_to_wikidata.ts
maxlath: If `createWdEntity` throws an error, `mergeEntities` will not be called, which is what we want, no? Did I get it wrong?

server/lib/utils/base.ts
maxlath: Yep, I knew this function's name would come up ^^

The `get` part is actually mimicking the likes of [lodash](https://lodash.com/docs/4.17.15#get)/underscore/backbone to access an object attribute's value. I still prefer it to the proposed alternatives. I think the `Optional` part of the name, or an equivalent signifier, should stay, as that's the purpose of that function: to not get a type error when trying to access an optional object attribute

jum-s (approved):

server/lib/utils/base.ts
maxlath: Discussion made obsolete by c6a5669 ^^'

PR #777, opened 2024-10-10
Allow the resolver to update Wikidata entities

maxlath:
Now that the server takes Wikidata editions into account (#753) and that it should soon be possible to move local entities  to Wikidata (#760), a missing brick was the possibility to perform the same kind of data enrichment we do on local entities directly on Wikidata entities, so that's what PR tries to do.

This is unfortunately hard to test as we don't mock edits on Wikidata, so Let's Test In Prod™ and see if further adjustments are needed.

One requirement was to make the resolver able to resolve to Wikidata editions, which is the other part of this PR.

~~PR based on #760~~

### DIFF
--- a/config/default.cjs
+++ b/config/default.cjs
@@ -217,12 +217,23 @@ const config = {
 
   deduplicateRequests: true,
 
+  // Keys for users OAuth
   // Doc: https://www.mediawiki.org/wiki/OAuth/For_Developers
   // Request tokens at
   // https://meta.wikimedia.org/wiki/Special:OAuthConsumerRegistration/propose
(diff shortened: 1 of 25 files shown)


maxlath:
Review delayed to 2025, to give time to milestones with a higher priority

maxlath:
Merging, as it solves a bug in prod (`updateResolvedEntry` unable to handle cases where the resolved edition is a Wikidata edition)

maxlath:
The newly generated Wikidata edits can be seen here https://www.wikidata.org/wiki/Special:Contributions/InventaireBot

PR #778, opened 2024-10-15
Entities merge requests

jum-s:
This PR solves #491 by allowing any logged-in user to ask for merging two entities of the same type (`human`, `work`, `serie`, `edition`, `publisher` and`collection`).

Before this PR, only users with a `dataadmin` role could merge entities. 

If a user submits a valid form from the `/entity/merge` menu, the server will either: 
  - merge the entities if some heuristics are valid (see `labelsMatchMergeByType`), 
  - or create a task document.

The other part (displaying tasks according to their `entitiesType`) is detailed in the client side PR (https://github.com/inventaire/inventaire-client/pull/514) since most of the work is done there


maxlath (changes requested):
First passage. One thing that struck me is that if, for example, for a given work we find matching labels with another work and matching labels between their authors, should both the work and the author be merged? Or shouldn't at least a task to merge the authors be created? But then why merge the works?

Something also came to mind about how we could make some other pre-merge checks, such as:
- checking for external identifiers conflicts: for instance, if author A has a BNF_ID=123 and author B has BNF_ID=456, that merge should probably be denied for non-dataadmins as cases where a national library gives several identifiers to the same author are fortunately rather rare. Chances are this would have been denied anyway by the non-matching work labels, but maybe there are several John Smith who wrote "Memoirs" and such, and a library identifier error would give a better context for the dataadmin investigating the task.
- deny merges between entities that have any claim linking to one another: this is already done to [remove merge suggestions on the client by the `getHomonymsEntities`/`parseSearchResultsToHomonyms`/`getRelationsUris` functions](https://github.com/inventaire/inventaire-client/blob/a157241/app/modules/entities/components/lib/homonym_deduplicates_helpers.ts#L45-L47)

This is what came to mind, but I'm also fine to go forward with this PR as is, regarding merge/task logic, and refine the heuristics later.

jum-s (commented):
First time ever that i have to do a pending review on top of a review, even though the process does not look different.

jum-s:
> One thing that struck me is that if, for example, for a given work we find matching labels with another work and matching labels between their authors, should both the work and the author be merged? Or shouldn't at least a task to merge the authors be created? But then why merge the works?

We agreed orally that graph resolution can happen in another PR.

>     * checking for external identifiers conflicts

done in 9fbffa4ad

>     * deny merges between entities that have any claim linking to one another

done in 904681c9e

It would be nice to not spend more developpement time (yours or mine) on this PR (except bug of course)

jum-s:
This branch has been rebased and autosquashed locally, waiting for review validation before force pushing it.

maxlath:
I had already some rebase conflict resolution recorded in `git rerere`, so I rebased and autosquash

maxlath (approved):

Issue #749, opened 2024-06-12
Allow subclasses of publishers


AutomCoding:
I got the invalid type error when adding Q10503565 to an edition.

I suspect it's because Gehrmans is only instance of subclass of a publisher. [The _publisher_ entity page](https://inventaire.io/entity/wd:Q10503565) redirects to [the _main subject_ page](https://inventaire.io/entity/wdt:P921-wd:Q10503565) so it isn't being recognized as a publisher.

maxlath:
Related issue: #414

maxlath:
Fixed by #784

Issue #414, opened 2020-08-17
check subclasses tree to type entities


maxlath:
## problem
ex: if someone creates a subclass of a literary genre such as [Q3739522](https://www.wikidata.org/wiki/Q3739522), and uses it as a genre (wdt:P136) value, it will reject the claim due to the "invalid entity type": the server is incapable to check the subclass tree to find that inherited type

## proposed solution 1
make get_entity_type an async function that can check inheritance

Pros:
- no need to add all the smallest types aliases to the [wikidata type aliases list](https://github.com/inventaire/inventaire/blob/master/server/lib/wikidata/aliases.js)

Cons:
- higher performance cost, due to those async operations, but that should be mostly limited to Wikidata entities

## proposed solution 2
generate [wikidata type aliases](https://github.com/inventaire/inventaire/blob/master/server/lib/wikidata/aliases.js) from a SPARQL query, periodically refreshed to integrate

Pros:
- no need for async just-in-time checks like in the previous solution

Cons:
- needs to be updated from time to time
- might need some special rules to work around things that could be considered subclasses of several types

jum-s:
[`get_entity_type`](https://github.com/inventaire/inventaire/search?q=get_entity_type&unscoped_q=get_entity_type) seems to be used when an entity creation or update happen. Those requests will be triggered only in case of unknown (non-aliased) P31 or P279 value, which greatly reduces this query occurrences to find subtypes.

Would the second proposition mean to store the subclass tree somewhere (ie. in the aliases list) ? If yes, then it doesnt look trivial as, for example, the number of subclasses for a big entity like [written work](https://query.wikidata.org/#SELECT%20%3Fs%20%3Flabel%20%7B%0A%20%20%3Fs%20wdt%3AP279%2a%20wd%3AQ47461344%20.%0A%20%20%3Fs%20rdfs%3Alabel%20%3Flabel%20filter%20%28lang%28%3Flabel%29%20%3D%20%22en%22%29.%0A%20%7D)(* operator :heart: ) is *quite* high. 

I dont know if there is a way to query all the superclasses of an entity instead.

maxlath:
[...]

PR #784, opened 2024-11-05
type entities using a list of possible wdt:P31 values extended by their subclasses

maxlath:
Addressing #414 #749 #787 #788

With this PR, what were formerly type aliases in `server/lib/wikidata/aliases.ts` become primary type aliases, which then `scripts/entities_extended_types_aliases/build_extended_aliases.ts` can use to fetch subclasses, to generate an extended list of type aliases in `server/assets/extended_types_aliases.json`

The first commits assume that building this extended list will be done at the server start and then long-term cached, but I then moved to a solution where the server only consumes a pre-built json file: it makes the server start less hazardous, but requires to then trigger freshness checks in a few places (pre-commit when meaningful, post-install, before entities indexation)


scripts/entities_extended_types_aliases/build_extended_aliases.ts
```diff
@@ -0,0 +1,125 @@
...
+async function getTypeExtendedAliases (type: PluralizedEntityType, sparqlRequests: string | string[]) {
+  const sparqlRequestsArray = sparqlRequests instanceof Array ? sparqlRequests : [ sparqlRequests ]
+  const hashCode = getHashCode(sparqlRequestsArray.join())
+  let extendedUris
+  try {
+    extendedUris = await cache_.get<WdEntityUri[]>({
+      // Use a a hash code in the key to bust outdated results when the query changes
+      key: `extended-wd-aliases:${type}:${hashCode}`,
+      fn: () => makeQueries(type, sparqlRequestsArray),
+      ttl: oneDay,
```
jum-s, line 51: Seems short, no ? Maybe this extended aliases is moving much more than I think, but `oneMonth` seems okay for a cache bust (doigt mouillé style)

scripts/entities_extended_types_aliases/build_extended_aliases.ts
```diff
@@ -0,0 +1,125 @@
...
+    typeExtendedAliases = difference(typeExtendedAliases, extendedTypesAliases.collections)
+  } else if (type === 'works') {
+    typeExtendedAliases = difference(typeExtendedAliases, [ ...extendedTypesAliases.series, ...extendedTypesAliases.collections ])
+  }
+  if (type === 'movements' || type === 'genres') {
+    typeExtendedAliases = dropOverlaps(type, typeExtendedAliases)
+  } else {
+    checkOverlaps(type, typeExtendedAliases)
+  }
+  extendedTypesAliases[type] = typeExtendedAliases.sort(byNumericId)
```
jum-s, line 36: What happens if something wrong happen (ie. a timeout) and `typeExtendedAliases` is actually empty (as it seems it could happen if `filteredIds` is not populated) ?
I cant a good suggestion to make sure `extended_types_aliases.json` is populated well, but hopefully you do?

scripts/githooks/pre-commit
```diff
@@ -4,3 +4,10 @@ set -eu
 
 echo -e "\e[0;30mstarting to lint...\e[0m"
 npm run lint-staged
+
+# Check entities extended types aliases freshness when at least one of the source file changes
+staged_entities_extended_types_aliases_files=$(git status --porcelain | grep --extended-regexp "^(A|M)" | grep --extended-regexp '.ts$' | sed --regexp-extended 's/^\w+\s+//' | grep -E '(extended_type|wikidata/aliases)' > /dev/null || echo "")
+echo "staged_entities_extended_types_aliases_files: $staged_entities_extended_types_aliases_files"
+if [ "$staged_entities_extended_types_aliases_files" != "" ] ; then
+  ./scripts/entities_extended_types_aliases/check_extended_aliases_freshness.ts 1
```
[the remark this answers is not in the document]
maxlath: Yes. That was useful for development, but not needed now anymore, done in f5d3b406f

server/assets/extended_types_aliases.json
```diff
@@ -0,0 +1,2233 @@
+{
+  "queriesHash": 336590939,
+  "extendedTypesAliases": {
```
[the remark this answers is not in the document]
maxlath: As said in 2778b2cc3 commit message

> the output of build_extended_aliases.ts might vary due to changes on Wikidata, tracking that file should help to get the same behavior across environments and machines, making it much more sane for debugging entity typing issues".

It shares some similarities with how `package-lock.json` is used to reproduce consistent environments npm-wise

scripts/entities_extended_types_aliases/build_extended_aliases.ts
maxlath: It's part of the motivation to track `extended_types_aliases.json`: if for some reason the build fails (because query.wikidata.org is down, or what not), it makes it still possible to use the last successful build until the build process can be fixed

server/assets/extended_types_aliases.json
```diff
@@ -0,0 +1,2233 @@
...
+      "wd:Q1149626",
+      "wd:Q1208380",
+      "wd:Q152559",
+      "wd:Q33289",
+      "wd:Q399495",
+      "wd:Q43091",
+      "wd:Q64362969",
+      "wd:Q839470"
+    ],
+    "meta": [
```
jum-s, line 2228: meta?

jum-s (approved):

Issue #271, opened 2019-02-09
merge the serie and work entity types


maxlath:
* some works are constituted of mutliple parts that might have editions for themselves #99 
* series might have editions of the whole serie at once

The trickiest part I can think of is that the client will have to recursively check for parts of any works now?

maxlath:
to fix once this is done:
* https://inventaire.io/entity/wd:Q1214163

maxlath:
Merging works and series might create some new issues: searching for series (to autocomplete a value for wdt:P179) for instance, would now return any work? So if you try to add the latest Asterix album to the Asterix series (wd:Q42490) would compete against all the other Asterix Albums? This might be solved by the series popularity score being higher than its parts, but it's still, in 99% of the cases, crowding results with undesired suggestsions.

Maybe:
- point 1 could be solved with a "serie" type lock, cf https://github.com/inventaire/inventaire/issues/99#issuecomment-1771442071
- point 2 could be solved by allowing wdt:P629 to have values of both "work" and "serie" types

maxlath:
A reason to keep the type distinction between "works known to have parts" aka "series" and "works not known to have parts" aka "works", is that in some cases it helps to prevent bad automatic edits. Example: see this error from the resolver:

```js
****** get_resolved_entry error ******
Error: invalid claim entity type: serie
    at error_.new (file:///home/admin/inventaire/server/lib/error/error.js:17:15)
    at verifyClaimEntityType (file:///home/admin/inventaire/server/controllers/entities/lib/validate_and_format_claim_value.js:76:21)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async Promise.all (index 1) {
  statusCode: 400,
  context: [
    'wd:Q3665579',
    entry: {
      edition: {
        isbn: '9784063722871',
        claims: {
          'wdt:P1476': [ 'チェーザレ : 破壊の創造者' ],
          'wdt:P648': [ 'OL27121119M' ],
          'wdt:P407': [ 'wd:Q5287' ],
          'wdt:P1104': [ 139 ],
          'wdt:P1680': [ 'Il creatore che ha distrutto' ],
          'wdt:P577': [ '2007' ],
          'wdt:P123': [ 'wd:Q726081' ]
        },
        labels: {},
        resolved: false
      },
      works: [
        {
          uri: 'wd:Q3665579',
          claims: { 'wdt:P648': [ 'OL19940640W' ] },
          labels: {},
          resolved: true
        }
[...]

Issue #99, opened 2017-06-30
some works are constituted of mutliple parts that might have editions for themselves


maxlath:
(just putting it here for later, but probably doesn't need to be solved right now)
In the current data model (see [entities map](https://inventaire.github.io/entities-map/), a series is constituted of works, which have editions, but some works might have sub-parts that could be considered works too, and have their own editions.
Example: [1Q84](https://www.wikidata.org/wiki/Q208971) (see discussion [here](https://www.wikidata.org/wiki/User_talk:Andreasmperu#revert_of_1Q84_P31) and [there](https://www.wikidata.org/wiki/Wikidata_talk:WikiProject_Books#should_a_novel_in_several_volumes_be_considered_a_book_or_a_book_series.3F)

And their might be editions of a whole serie. The current data model is thus too rigid, and might need to consider series and works as one unique abstract layer that might have sub-parts of the same nature as well as direct editions.

maxlath:
Ideally, we would need to let the work concept have the **possibility** (while, in most cases it wouldn't be needed) to span from whole series to short stories or chapters, letting editions the possibility to precisely pick from which work they are an edition (or [composite edition](https://github.com/inventaire/inventaire/issues/93)) of: the need for a work to be created would arise from the existence of an edition needing this level of precision. "Work" would be all at once the atom and the complex molecule by which we can divide writings, potentially recursively containing several elements of the same nature. The same logic could later apply to other kinds of products/goods

maxlath:
This kind of issue could be solved with https://github.com/inventaire/inventaire/issues/181 by locking this kind of rare exception to be considered of type "serie"(?)

PR #785, opened 2024-11-05
implement entities type lock

maxlath:
Sometimes, Wikidata's attempt to describe reality with fidelity, while also accommodating its own priorities (i.e. being primarily at the service of Wikimedia projects), leads to our entity typing system astray. For instance, the author of  [1Q84](https://inventaire.io/entity/wd:Q208971) insists that it's not a book serie, but a single book in 3 volumes. Across the world, those 3 volumes are sometimes edited together, or 2+1, all separately. Within Inventaire entities ontology, this would be best modeled as a series made of 3 works, but our current typing logic, based on [Wikidata P31 values](https://www.wikidata.org/wiki/Q208971#P31) finds [`wd:Q208971`](https://inventaire.io/entity/Q208971) to be a work and not a serie.

This PR implement a way to override that typing by setting a local claim, using the ad hoc local property "entity type lock" (`invp:P3`). This thus heavily relies on the work done in https://github.com/inventaire/inventaire/pull/753

This would help bring a much-needed sanity to some otherwise nerve wrecking edge cases, but at the cost of more of a persistent divergence from Wikidata: this should thus be kept to a minimum (that is, after the options to fix the issue directly on Wikidata has been explored). Consequently, the property would initially be editable only by admins, but could later be extended to dataadmins.

Related issues: #99 #271


server/controllers/entities/lib/create_wd_entity.ts
```diff
@@ -16,7 +16,7 @@ import type { SimplifiedQualifiers, WikimediaLanguageCode } from 'wikibase-sdk'
 
 const wmLanguageCodeByWdId = requireJson('wikidata-lang/mappings/wm_code_by_wd_id.json')
 
-const allowlistedEntityTypes = [ 'work', 'serie', 'human', 'publisher', 'collection', 'edition' ]
+export const allowlistedEntityTypes = [ 'work', 'serie', 'human', 'publisher', 'collection', 'edition' ]
```
jum-s, line 19: Why not reusing `allLocallyEditedEntitiesTypes`? It does feel okay to infer "editable entities types" === "creatable entities type". Too much DRYing? I can see our future self reusing `allowlistedEntityTypes` within other scopes than this current 'entity creation' scope, without paying attention

server/controllers/entities/lib/properties/properties_config_bases.ts
```diff
@@ -77,3 +79,14 @@ export const imageHash = {
...
 } as const
+
+export const entityType = {
+  datatype: 'entity-type',
+  primitiveType: 'string',
+  format: trim,
+  validate: ({ value }) => {
+    return allLocallyEditedEntitiesTypes.includes(value)
+  },
+  adminEditOnly: true,
```
jum-s, line 90: It would be more adaptable to directly have the access level written here:
```suggestion
  editAccessLevel: 'admin',
```
This would need to transform PR occurrences of `userIsAdmin: userHasAdminRole(user)` -> `userAccessLevel: getUserAccessLevels(user)`

server/controllers/entities/lib/validate_entity.ts
```diff
@@ -14,26 +16,31 @@ const allowlistedTypes = allLocallyEditedEntitiesTypes
...
+  const isLocalLayer = isLocalEntityLayer(entity)
+  let type
+  if (isLocalLayer) {
+    const remoteEntityUri = getFirstClaimValue(claims, 'invp:P1')
+    const remoteEntity = await getEntityByUri({ uri: remoteEntityUri })
+    type = remoteEntity.type
+  } else {
+    type = getValueType(claims)
+    validateValueType(type, claims['wdt:P31'] as WdEntityUri[])
+    validateLabels(labels, type)
```
jum-s, line 41: Is there a reason why local layer entities are now excluded from those validations: `validateValueType` and `validateLabels`? Admins could be mistaken too when they're typing.

server/controllers/user/lib/user.ts
```diff
@@ -107,6 +107,11 @@ export const addUserRole = (userId: UserId, role: UserRole) => db.update(userId,
 
 export const removeUserRole = (userId: UserId, role: UserRole) => db.update(userId, removeUserDocRole(role))
 
+export function userHasAdminRole (user: User) {
```
jum-s, line 110: I'm not much in favour of this function, It would be preferable to use `getUserAccessLevels`

server/models/validations/validate_required_properties_values.ts
```diff
@@ -1,22 +1,31 @@
...
-    validateControlledPropertiesClaimsPerType[type](claims)
-  }
-
-  checkedProperties = checkedProperties || (Object.keys(claims) as PropertyUri[])
-
-  checkedProperties.forEach(property => {
+  objectKeys(claims).forEach(property => {
     validateUniqueValue(property, claims[property])
+    if (isLocalLayer) {
+      rejectRemotePropertiesOnLocalLayer(property, doc)
```
jum-s, line 25: I understand `reject` as the opposite of `filter`, but this naming actually infers `reject` as the opposite of `validate`. I would be in favour of only using the verb `validate` for such functions.

```suggestion
      validateLocalLayerWithoutRemoteProperties(property, doc)
```

server/models/validations/validate_required_properties_values.ts
```diff
@@ -1,22 +1,31 @@
...
+    } else {
+      rejectRemoteEntityOnlyProperty(property, doc)
```
jum-s, line 27:
```suggestion
      validateLocalEntityWithoutRemoteProperty(property, doc)
```

tests/api/entities/type_lock.test.ts
```diff
@@ -0,0 +1,46 @@
...
+    const uri = await getSomeWdEditionUri()
+    await addClaim({ uri, property: 'invp:P3', value: 'work' })
+    .then(shouldNotBeCalled)
+    .catch(err => {
+      err.statusCode.should.equal(403)
+      err.body.status_verbose.should.equal("editing property requires admin's rights")
+    })
+  })
+
+  it('should reject invalid types', async () => {
```
jum-s, line 18: What about having a test to assert that an `invp:P3` cannot be added if it is the same as the current type:

```
  it('should reject when typed-locked entities type is the same as the current entity type', async () => {
```

server/controllers/entities/lib/create_wd_entity.ts
maxlath: I was making for `create_wd_entity.ts` to have its own list when that was only a subset of the possible local types, but now that's obsolete, so it can be fully removed: 32c74e220

server/controllers/entities/lib/properties/properties_config_bases.ts
maxlath: It might be YAGNI, but I gave it a try in 66db7d818

server/controllers/entities/lib/validate_entity.ts
maxlath: Because at this point, the local layer is only that: a local layer, without the remote entities labels and claims, in particular its `wdt:P31` claims

server/models/validations/validate_required_properties_values.ts
maxlath: Done in 70fda245c

server/models/validations/validate_required_properties_values.ts
maxlath: Done someting in that direction in 70fda24

tests/api/entities/type_lock.test.ts
maxlath: There might be cases where that would be annoying: imagine that an item identified as having an unstable type, because it's has two `wdt:P31` values

Issue #574, opened 2021-12-08
create several posts with one book each on AP when adding books


roipoussiere:
When adding several books on inventiare.io, a post with multiples books is created on the fediverse:

![image](https://user-images.githubusercontent.com/1665542/145193667-470bfd2c-3955-4254-b0c2-7e0b185540f2.png)

I think it could be interesting to create several posts with one book each: this way fediverse users could add a comment on one specific book.

I'm not talking about public book reviews (#185), but about commenting the owning of a book by a specific user. Comments could be something like "oh I liked this one, it remembered me our travel last year" or "hey, can I borrow you this one in two weeks?".

Then, two related features could be added in the future:
- add the possibility to display these comments on the inventaire.io website;
- add the possibility to directly add this kind of comment on the inventaire.io website.

maxlath:
the grouping was done with large imports in mind, like when someone would import 100 books at once from a pre-existing list, but I see how that can be undesired for smaller activities. Some ideas:
* there could be a default threshold under which several items create several activities
* that threshold could be customized by the user

roipoussiere:
What is the problem to create 100 posts? It could be done in a progressive way, for instance by waiting few seconds between each post.

When grouping posts, I don't understand how a fediverse user could comment on the owning of a book.

jum-s:
Allow more control on the published activities, in order to be able to disable activity pooling, there will be a need to publish one post per item added to the inventory.

A setting config may be accessible to users to enable/disable this feature.

For the moment, max and i agreed to have this feature disabled by default (according to MoU call today)

PR #802, opened 2024-12-09
Disable published activities pooling

jum-s:
This is an attempt to "unpool" recent activities.

While having to deal with past decisions (storing in db `since` and `until` inside one activity, seems rigid to our current needs...), this implementation is still relying on those stored activities, to avoid having to re-write/touch on the current activities stored in db.

Choosing this implementation has the advantages of fully reusing existing functions (refactoring for directly formatting items could be done later) 

To-do later: delete any pooling possibilities, and stop storing `since` and `then` activities

Solves #574
Client PR #526
To be merged after #801


maxlath:
[non-blocking consideration]
I was wondering if `user.poolActivities` should rather be `user.settings.activities.pool`, to keep this kind of user-updatable parameters together, but there are other settings not following that convention (`fediversable`, `summaryPeriodicity`, etc) so maybe we are stuck with this inconsistency for now

server/controllers/activitypub/lib/radio_hooks.ts
```diff
@@ -42,11 +42,12 @@ const createDebouncedActivity = ({ userId, shelfId }: createActivityParams) => a
 }
 
 async function _createDebouncedActivity ({ userId, shelfId }: createActivityParams) {
-  let name, user
+  let name, user, poolActivities
   if (userId) {
     delete debouncedActivities[userId]
     user = await getUserById(userId) as User
     if (!user.fediversable) return
+    ({ poolActivities } = user)
```
maxlath, line 50: [non-blocking]
Please prefix lines starting with `(` with a `;`, I had to check if this was not interpreted as `return ({ poolActivities } = user)` (It doesn't). A `;` let's us see the danger by its sheer ugliness, and prevents a bug if the line above changes.
```suggestion
    ;({ poolActivities } = user)
```

server/controllers/activitypub/lib/radio_hooks.ts
```diff
@@ -68,12 +70,11 @@ async function _createDebouncedActivity ({ userId, shelfId }: createActivityPara
     actor: { name },
     object: { items: { since, until: Date.now() } },
```
maxlath, line 71: If `user.poolActivities === true`, no need to create this activity doc: the item doc itself has all we need (just like entity activities are built from patch docs, without needing to store an activity doc)

maxlath (changes requested):

server/controllers/activitypub/lib/radio_hooks.ts
```diff
@@ -42,11 +42,12 @@ const createDebouncedActivity = ({ userId, shelfId }: createActivityParams) => a
 }
 
 async function _createDebouncedActivity ({ userId, shelfId }: createActivityParams) {
```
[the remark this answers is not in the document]
jum-s: I don't get it, for me, your comment says:

`debounce === pool`
so `if (pool === true)` 
then `debounce` should be `false` 

Can you see the contradiction I see?

If we pool, activities are debounced, no? As it was the case before this PR. Or I don't understand at all what's going on here?

server/controllers/activitypub/lib/radio_hooks.ts
jum-s: if it was me i would prevent using this variable definition, and force into `poolActivities = user.poolActivities`
I added `;` in c258dceb4 (exceptionnally not in a fixup, but amended)

server/controllers/activitypub/lib/radio_hooks.ts
jum-s: I explored how to do what you suggest. After a few hours of refactoring, i think its not worth it anymore.
Because `poolActivities` is used quite deeply in the current code: at the very formatting of the activity (`formatUserItemsActivities`) and it's used by the `outbox` endpoint, the `activity` endpoint and the `postActivity` function. 
So extracting `poolActivities` (necessary to pass directly an `items` array to `formatUserItemsActivities`, what you say with "the item doc itself has all we need") leaves an early `if` which then have functions that almost look like each other (ie.
- `formatUserItemsActivities({ activitiesDocs, user })` and 
- `formatUserItemsActivities({ items, user })` 
)
This duck typing is more difficult to read, and introduces different behaviors using different objects (`items` & `activities`), it feels it will be even harder to refactor later IMO. (Plus, all the refactoring have to be done with the `shelves` part too)

Your comment/diagnosis is legit, but the intervention is trickier than expected. The main reason behind is that we now want to keep both behaviors available (pooling and not pooling). The fact that it is now unpooled by default, makes it easy to check later is this pooling feature is actually activated by people. What about leave the PR like this, wait to see if the pooling feature is actually used, and then decide to remove or keeping it once we figured it out ?

Note: after the exploration, i ended up with the feeling that if the pooling feature is here to not spam other instances, what about only having pooling when sending create activities, but not on `outbox` and `activity` endpoints ? This would induce to remove the possibility to send request to the `activity` endpoint and directly build the link in the pooled activity only with the first item created.
Note2: The more i think about it, the more I think wa should just get rid of the complexity by removing pooling possibility all together

server/controllers/activitypub/lib/radio_hooks.ts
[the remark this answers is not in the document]
maxlath: I messed up the logic in my comment, sorry, see edited version

Issue #719, opened 2024-01-28
Deliver activities within non-blocking jobs


maxlath:
so that a failing delivery can be retried later

maxlath:
Done in https://github.com/inventaire/inventaire/pull/803

PR #803, opened 2024-12-09
Activity jobs delivery

jum-s:
This PR allows the server to retry posting activities on remote instances after failing. 
It relies on function `initJobQueue` and leveldb in the background to achieve this.

Solves #719
To be merged after https://github.com/inventaire/inventaire/pull/801

### DIFF
--- a/server/controllers/activitypub/lib/post_activity.ts
+++ b/server/controllers/activitypub/lib/post_activity.ts
@@ -83,7 +83,7 @@ export async function fetchInboxUri ({ actorUri, activity }: { actorUri: Absolut
 
 async function buildAudience (activity, inboxUrisByBodyTos) {
   const actorUri: AbsoluteUrl = activity.actor.uri
-  const inboxUri = await fetchInboxUri({ actorUri, activity })
+  const inboxUri = await fetchInboxUri({ actorUri, activity }) as AbsoluteUrl
   if (inboxUri) {
     if (inboxUrisByBodyTos[inboxUri]) {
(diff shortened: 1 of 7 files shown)


server/controllers/activitypub/lib/post_activity.ts
```diff
@@ -28,19 +28,8 @@ export async function postActivity ({ actorName, inboxUri, bodyTo, activity }: {
...
-      timeout,
-      parseJson: false,
-      retryOnceOnError: true,
-    })
-  } catch (err) {
-    err.context = err.context || {}
-    Object.assign(err.context, { inboxUri, activity })
-    logError(err, 'Posting activity to inbox failed')
-  }
+  await postActivityQueue.push({ inboxUri, postHeaders, body, activity })
```
maxlath, line 31: Aren't signed requests time sensitive? If the job takes too much time to be done (because a lot is going on or we later decide to `waitForCPUsLoadToBeBelow(x)`, I'm afraid some signed requests might fail. It seems that could easily be fixed by having the job work also do the part that calls `signRequest`, by passing all `postActivity` arguments as job payload data

maxlath (commented):
NB: I pushed some commits to this branch

server/controllers/activitypub/lib/post_activity.ts
jum-s: Right, I think it's 30 seconds for mastodon, which is quite some time for a job to be executed. But hey, indeed the fix is easy (done in 87d3624a5)

PR #806, opened 2024-12-18
Landing map

jum-s:
This is mainly opening a `items/by-bbox` endpoint in order to serve the client landing page map [#527](https://github.com/inventaire/inventaire-client/pull/527)

This seems cleaner than going through the existing `by-position` endpoint, since the client is displaying the exact books of users displayed on the map, while `by-position` would serve every item within a range/radius, ending potentially displaying items which do not belong to anyone on the map.

### DIFF
--- a/server/controllers/items/by_bbox.ts
+++ b/server/controllers/items/by_bbox.ts
@@ -1,0 +1,29 @@
+import { property } from 'lodash-es'
+import { filterMaximumItemsPerOwner } from '#controllers/items/lib/filter_maximum_items_per_owner'
+import { getItemsByUsers } from '#controllers/items/lib/get_items_by_users'
+import { getUsersByBbox } from '#controllers/user/lib/user'
+import type { SanitizedParameters } from '#types/controllers_input_sanitization_parameters'
+import type { Item } from '#types/item'
+import type { User, UserId } from '#types/user'
+
+const sanitization = {
+  bbox: {},
+  limit: {
+    default: 15,
+  },
+  lang: {
+    default: 'en',
+  },
+}
+
+async function controller (params: SanitizedParameters) {
+  const { bbox, limit, lang } = params
+  const foundUsers: User[] = await getUsersByBbox(bbox)
+  const usersIds: UserId[] = foundUsers.map(property('_id'))
+  const { items, users }: { items: Item[], users: User[] } = await getItemsByUsers({ ...params, usersIds })
+
+  const filteredItems = filterMaximumItemsPerOwner(items, lang, limit)
+  return { items: filteredItems, users }
+}
+
+export default { sanitization, controller }
--- a/server/controllers/items/lib/filter_maximum_items_per_owner.ts
+++ b/server/controllers/items/lib/filter_maximum_items_per_owner.ts
@@ -1,0 +1,21 @@
+export function filterMaximumItemsPerOwner (items, lang, limit) {
+  const filteredItems = []
+  const discardedItems = []
+  const itemsCountByOwner = {}
+
+  for (const item of items) {
+    if (filteredItems.length === limit) return filteredItems
+    itemsCountByOwner[item.owner] ??= 0
+    if ((item.snapshot['entity:lang'] === lang) && (itemsCountByOwner[item.owner] < 3)) {
+      itemsCountByOwner[item.owner]++
+      filteredItems.push(item)
+    } else {
+      discardedItems.push(item)
+    }
+  }
+
+  const missingItemsCount = limit - filteredItems.length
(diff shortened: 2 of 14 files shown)


maxlath:
Rebased

server/controllers/items/lib/filter_maximum_items_per_owner.ts
```diff
@@ -0,0 +1,23 @@
+import type { SerializedItem } from '#types/item'
+
+export function filterMaximumItemsPerOwner (items: SerializedItem[], lang: string, limit: number) {
+  const filteredItems = []
+  const discardedItems = []
+  const itemsCountByOwner = {}
+
+  for (const item of items) {
+    if (filteredItems.length === limit) return filteredItems
+    if (itemsCountByOwner[item.owner] == null) { itemsCountByOwner[item.owner] = 0 }
```
maxlath, line 10:
```suggestion
   itemsCountByOwner[item.owner] ??= 0
```

server/controllers/items/by_bbox.ts
```diff
@@ -0,0 +1,29 @@
...
+  },
+  lang: {
+    default: 'en',
+  },
+}
+
+async function controller (params: SanitizedParameters) {
+  const { bbox, limit, lang } = params
+  const foundUsers: User[] = await getUsersByBbox(bbox)
+  const usersIds: UserId[] = foundUsers.map(property('_id'))
```
maxlath, line 22: Without a limit on the bbox size, this could end up being a lof of users. The bbox is constrained by the client, but I don't think it would hurt to `.slice(0, 500)`, given that this endpoint also slices the returned items (not aiming for exhaustivity). A more sophisticated version could sort the users before slicing (for instance on the `user.updated` attribute to prioritize users more likely to be active users) .

tests/api/items/get_by_bbox.test.ts
```diff
@@ -0,0 +1,31 @@
...
+import { fixedEncodeURIComponent } from '#lib/utils/url'
+import { waitForIndexation } from '#tests/api/utils/search'
+import { publicReq } from '#tests/api/utils/utils'
+import type { LatLng } from '#types/common'
+
+const endpoint = '/api/items?action=by-bbox'
+
+const userPosition: LatLng = getRandomPosition()
+const [ lat, lng ] = userPosition
+const geolocatedUserPromise = createUser({ position: userPosition })
```
maxlath, line 14: Please avoid creating fixtures outside of test functions, as it means those are created even if the test are ignored/xitted, which adds noise to the tests logs. If sharing fixtures between tests is needed, such variable can be initialized from a `before` function, see [example](https://github.com/inventaire/inventaire/blob/377c5bf/tests/api/entities/serie_parts.test.ts#L11)

tests/api/items/get_by_bbox.test.ts
```diff
@@ -0,0 +1,31 @@
...
+const bbox: string = fixedEncodeURIComponent(JSON.stringify([
+  lng - 1, // minLng
+  lat - 1, // minLat
+  lng + 1, // maxLng
+  lat + 1, // maxLat
+]))
+
+describe('items:public:by-bbox', () => {
+  it('should get anonymized public items nearby', async () => {
```
maxlath, line 23: what's "anonymized" about those items?

tests/api/items/get_by_bbox.test.ts
```diff
@@ -0,0 +1,31 @@
...
+    const user = await geolocatedUserPromise
+    await waitForIndexation('users', user._id)
+    const item = await createItem(user)
+    const { items } = await publicReq('get', `${endpoint}&bbox=${bbox}&strict-range=true`)
```
maxlath, line 27:
```suggestion
    const { items } = await publicReq('get', `${endpoint}&bbox=${bbox}`)
```

jum-s: Done in 32114a27e

tests/api/items/get_by_bbox.test.ts
jum-s: Legacy from former implementation, removed in 32114a27e

tests/api/items/get_by_bbox.test.ts
jum-s: Thanks for the example. Done in 32114a27e

server/controllers/items/by_bbox.ts
jum-s: Done in 2b21d52b2

server/controllers/items/lib/filter_maximum_items_per_owner.ts
jum-s: Done in 8c7816dd9

PR #796, opened 2024-11-27
Allow a server instance to edit entities from a remote instance

maxlath:
Addressing https://github.com/inventaire/inventaire/issues/549 by letting server instances proxy requests to entities-related endpoints (1) to a remote instance set in config (`config.federation.remoteEntitiesOrigin`). Those proxied requests are made as signed HTTP requests with a `x-remote-user` header to pass an anonymizable user id (2), giving the remote server the guaranty that the request comes from a certain domain, and trusting that server on who the user is.

(1):  `/api/entities`, `/api/tasks`, and `/api/data` endpoint groups, as well as `/api/images?action=convert-url` and `/api/images?action=upload`

(2) That anonymizable user id is a random [UUID](https://en.wikipedia.org/wiki/Uuid) set on the user doc to allow identifying patches while leaving the decision to the user on whether the patch author username and profile pic should be public (depending on the user value for `user.settings.contributions.anonymize`).

Client PR: https://github.com/inventaire/inventaire-client/pull/528

TODO post-merge:
- [ ] set `anonymizableId` on all user docs, that can be used to generate the user `acct`
- [ ] convert patches `user` attribute from user ids to `acct`
- [ ] convert tasks `reporters` from user ids to account uris `acct`
- [ ] update prod config (ex: `config.name` => `config.instanceName`)

TODO later:
- implement webfinger for `anonymizableId` based user account uris (returning only what `user.settings.contributions.anonymize` allows, like for `/api/users?action=by-accts`)

Issue #722, opened 2024-02-01
Implement remote instance entities caching strategy 


jum-s:
A caching strategy between servers provides a few advantages (non-exhaustive): 
- reduce server2server number of requests,
- improve instance anonymization,
- avoids hitting inventaire.io's cache too much
 
Invalidation implementations: 
The make-it-work version would trigger a refresh from instance server at periodical time (on the model of client cache refresh (TTL based)). 
A make-it-right version may rely on pushing notification updates from inventaire.io, to keep up to date data across instances, while reducing requests between servers. This would required to store domains who already requested an entity(?)

PR #368, opened 2020-02-25
[WIP] add inventory view cache

maxlath:
adds caching per user and access level (`public`, `network`, `private`) to limit the amount of cached data while maximizing the cache hits:
* inventory-view request for a single user => possibly fully cached
* inventory-view request for a group => gets possibly cached individual inventory views from users, and aggregates it (this last operation isn't cached to limit required cache busts)

TODOs: bust cache on every events that might change the inventory view data:
* [x] item created/updated/deleted => bust affected caches (possibly several, depending on the listing (aka visibility level)
* [x] edition entity update on work (wdt:P629) claims
* [x] work entity update on author (wdt:P50, +), genre (wdt:P136), subject (wdt:P921) claims

PR #812, opened 2025-02-23
Add cache to federated requests

maxlath:
Addressing #722 

## Server instance caches

Fill the remaining holes following #796 and #810 so that displaying inventories on a federated instance can be done with a as few requests to the entities host instance as possible and reasonable. The main challenge being of course cache invalidation, and in particular the issue of having different part of the app displaying variously fresh data. In that regard, I'm quite happy with the solution proposed in #810, to let requests to `/api/entities?action=by-uris` hit the entities host server, and to use the response to call `updateEntitiesRevisionsCache` and invalid the different caches, based on entities' revision numbers.

The main cache in the social core remains items snapshots: doing the heavy lifting to keep display inventory data while keeping requests to entities minimal.

This PR introduces a new generalist cache used by the `getRemoteEntitiesByUris` function, which will be on a hot path, due to its use by the `/api/items?action=inventory-view` controller (and in absence of an inventory view cache, cf #368).

Both the items snapshots and the `getRemoteEntitiesByUris` cache will be invalidated via the revisions cache (by listening to the `entity:change` event).

## Browser cache
Additionally to the server instance caches, this PR takes advantage from the [`if-none-match` HTTP header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-None-Match) sent by the browser to reduce data exchanges between the federated server and the entities host server: if the [ETag](https://en.wikipedia.org/wiki/HTTP_ETag) sent by browser and passed by the federated server proxy to the entities host server matches the response content, that server will just return a `304` and an empty body, which will be repeated by the federated server to the browser.

## Considerations on the future of decentralized caching
We should probably refrain from heavily caching entities to not frustrates users to much: already with just inventaire.io and wikidata.org, we get users complaining that the data sometimes changes from one page to the other (which typically happens when going from an entity layout to the entity editor, as the later triggers a request with `refresh=true`). Let's see what pain points emerge from the current implementation


tests/api/entities/federated_entities.test.ts
```diff
@@ -1,15 +1,54 @@
...
     .catch(err => {
       err.statusCode.should.equal(400)
       err.body.status_verbose.should.equal('This endpoint is closed in federated mode')
     })
   })
+
+  describe('proxied requests', () => {
+    // xitted, as it's not actually testing if the request between
+    // the remote entities origin and the federated server was also a 304
+    xit("should return a 304 if the response didn't change", async () => {
```
jum-s, line 23: Can the comment say why it's not actually testing the request ?

tests/api/entities/federated_entities.test.ts
```diff
@@ -1,15 +1,54 @@
...
 describe('entities:federated mode', () => {
-  it('should close endpoints not yet implemented in federated mode', async function () {
-    if (!federatedMode) this.skip()
+  before(function () { if (!federatedMode) this.skip() })
+
+  it('should close endpoints not yet implemented in federated mode', async () => {
     await authReq('put', '/api/entities?action=revert-merge')
     .then(shouldNotBeCalled)
     .catch(err => {
       err.statusCode.should.equal(400)
```
jum-s, line 15: This is red for my federated tests `AssertionError: expected 403 to be 400`

```
  context: {
    roles: [ 'public', 'authentified' ],
    requiredAccessLevel: 'dataadmin'
  },
  emitter: [
    'Error: unauthorized api access',
...
  ],
  user: { _id: '9f115db535c1b349a9cd1d0354000bdc', username: 'Halcombed9Iz' },
  headers: {
    'user-agent': 'federated-tests-federated-mocha',
    'content-length': '0'
  }
```

jum-s (approved):

